PR #7, opened 2021-11-12
feat: option to comment out lines in the bleep.json file

ingarabr:
JSON does not support to comment out lines. Filter out lines to make it a bit easier to work with the build configuration.

### DIFF
--- a/bleep-core/src/main/scala/bleep/model.scala
+++ b/bleep-core/src/main/scala/bleep/model.scala
@@ -593,5 +593,5 @@ object model {
   }
 
   def parseBuild(json: String): Either[Error, Build] =
-    parser.decode[model.Build](json)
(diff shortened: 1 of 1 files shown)


oyvindberg:
:fireworks:

sideeffffect:
Aren't you worried that now the format of `bleep.json` isn't JSON but something else? That can come with number or issues, sooner or later.

Sure, JSON is very limited (intentionally so, AFAIK) and there is surely a demand for more something more feature-full. Wouldn't adding support for YAML be a better choice here?

oyvindberg:
Not terribly so, honestly. I like the flexibility of commenting out things for development/debugging. 

After the fact comments are very easily stripped out. In fact bleep will have several build-rewriting commands (normalize json, bump dependencies, add cross build, etc) which all will drop the comments.

That said, im very open to looking into different formats, so yaml could possibly still be explored. See also #15

PR #10, opened 2021-11-14
feat: use decline to parse command line inputs

ingarabr:
It's not perfect but it gives a good baseline to continue working on it.


oyvindberg:
cool, this will do fine for now.

if you have any more time to spare we really need awesome tab completion for this, see #11

PR #16, opened 2021-11-17
Bsp support

oyvindberg:
BSP spike. Successfully starts a bloop server, connect through tcp, run compile, stop server

- import a ton of code from scala-cli, which has already solved the problem
- inherits a bunch of reflection problems when creating native image.
- should probably replace bsp4j with bsp4s
- move our bloop files to .bleep/.bloop


oyvindberg:
good enough to merge this first PR. and big enough. mostly big enough.

PR #17, opened 2021-11-18
Support a richer model for dependencies

hamnis:
We now support a potential richer model for when we need to include classifiers, exclusions or optionallity

```json
"dependencies":  [
  "org.http4s::http4s-client:0.23.1",
  { 
    "module": "org.http4s::http4s-ember-client:0.23.1",
    "excludes": ["org.http4s::http4s"],
    "optional": false,
    "classifier": "sources"
  }
]
```


oyvindberg:
Nice! Does this support provided scope as well? Or do you know how that works?

hamnis:
Scopes are configurations, so that is easy to add.

Issue #1, opened 2021-10-29
Specify JVM in bleep.json


oyvindberg:
Coursier can manage JVMs. As such a bleep file should be able to just specify one, and it should be downloaded and used automatically.

There is already functionality for this in the ported native image plugin.

PR #34, opened 2021-11-20
Work on interface

oyvindberg:


bleep/src/main/scala/bleep/BleepCommandRemote.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def chosenTargets(started: Started): util.List[BuildTargetIdentifier] =
+    started.chosenProjects.map(projectName => targetId(started.buildPaths, projectName)).asJava
+
+  def runWithServer(started: Started, bloop: BloopServer): Unit
+
+  override final def run(): IO[ExitCode] =
+    runWithEnv { started: Started =>
+      val bloopRifleConfig = new BloopSetup(
+        "/usr/bin/java",
```
hamnis, line 25: $JAVA_HOME/bin/java, the we can fallback to this?

oyvindberg: Or even better https://github.com/oyvindberg/bleep/issues/1 . Up for grabs if you wanna try, otherwise I'll get to it in some days

PR #44, opened 2021-12-10
Use graalvm 21

hamnis:


oyvindberg:
maybe some inspiration here , https://github.com/VirtusLab/scala-cli/blob/master/.github/workflows/ci.yml .

oyvindberg:
🚀

PR #76, opened 2022-02-04
Use bloop-rifle from scala cli

hamnis:


oyvindberg:
Great, thanks! 

Awesome to not vendor that code when we don't have to.

We'll now need to add some way to configure bloop, because everything gets a lot slower when we always terminate it after every command.

hamnis:
a bloop subcommand section here might be useful. 
Detection of an already running bloop server should be done as well, we now always start a new one.

PR #81, opened 2022-02-07
Improve dependency culling

oyvindberg:
This goes some way towards solving this problem. The problem with the import as it was was two-fold:

### Exact version numbers 
- we only culled dependencies which were transitive from other dependencies with that same exact version number. Now the check only keeps a dependency if it has a higher version number than any transitively inherited dependencies.

### parsing raw POM files

These may use the `dependencyManagement` feature of POMs, along with parent POM(s) (note how most of the unnecessary were java libraries). Furthermore, they may use templating. `${project.groupId}` with friends. As a result, we got a bunch of dependencies with empty version numbers.
I solved this by delegating everything to `coursier.Resolve`. I won't do that for *all* dependencies, because it is very, very slow. 

## Status

~~To my eyes the code looks good, but it currently produces wrong results (it removes too many dependencies).
As such we'll need #3 to go forward here. if we test-import a bunch of complex builds we can make a property test that the full set of transitive dependencies should be the same before and after import~~ Seems to work well.

There is still a problem with `scala-reflect` which I'm guessing is caused by missing implementation of `provided` but I'll look into that later


bleep/src/main/scala/bleep/internal/importBloopFilesFromSbt.scala
```diff
@@ -291,7 +298,12 @@ object importBloopFilesFromSbt {
                 logger.warn(s"Couldn't determine dependencies for $mod: $errMsg")
                 Nil
               case Right(project) =>
-                project.dependencies
+                // this means the version is set in a parent POM. there may also be templating involved. Let's defer to coursier for that.
+                if (project.dependencies.exists { case (_, dep) => dep.version == "" }) {
+                  // slow path
+                  val resolved = Resolve().addDependencies(Dependency(project.module, project.version)).run()
```
hamnis, line 304: Can we resolve everything in one go?

def of(...): Either[NeedsResolving, ParsedDependency]
then after we can collect all lefts and resolve those?

oyvindberg: maybe! I'll try :)

oyvindberg:
@hamnis look at how extensively we can track all changes in the dependency graphs with the new test suite. Now we have to tools needed to land this

PR #83, opened 2022-02-11
Setup integration tests

oyvindberg:
I think this is surprisingly cool. Snapshot tests absolutely rock

## Setup / add an integration test for an sbt build
- First we add the build as a git submodule. This is not really needed, but I want to future proof for running build tests. It's not really needed in that case either, because we could always clone on demand. Anyhow, git submodules it is for now.
- Then, the test framework runs sbt to generate bloop files. this is typically done by the developer when adding the given build, or when updating it to a newer version. we do not run sbt during integration tests because it is so slow.  
- since bloop files are tied to a computer, we generalize them by templating in replacements for current directory and so on. This way we can run the tests on any computer

## Actually running tests
- first we template it appropriate values for directories
- if we're missing any dependencies on the current machine we resolve everything for a given project. this is done because the import needs pom files to work. it'll also be necessary for compiling and/or running tests later
- then we run a full import with templating (irritating name collision! the templating that is used to extend projects) and so on, and generate new bloop files based on that
- both the generated bleep.json and the bloop files are checked as snapshot tests. in CI they are required to match the checked in versions, locally they will just be overwritten

## What is tested?
The entire import flow, templating, dependency culling, generating bloop files. basically all the complicated parts.

## todo:
- ~~compare classpath, (re-)source directories, platform and any other useful parts of the (input, output) pairs of bloop files~~ Some checks are enabled now, many will need additional bugfixes to run. more work in this direction soon


bleep-core/src/main/scala/bleep/internal/FileUtils.scala
```diff
@@ -1,12 +1,82 @@
...
+      val bytes = content.getBytes(StandardCharsets.UTF_8)
+      val synced = if (soft) softWriteBytes(file, bytes) else writeBytes(file, bytes)
+      ret(file) = synced
+    }
+
+    ret.toMap
+  }
+
+  def softWriteBytes(path: Path, newContent: Array[Byte]): Synced =
+    (if (FileUtils.exists(path)) Some(Files.readAllBytes(path)) else None) match {
```
hamnis, line 62: `Option.when(FileUtils.exits(path)(FIles.readAllBytes(path)`

oyvindberg:
also todo: move bleep cache inside the coursier cache folder for the integration tests, so CI don't have to resolve each time

PR #97, opened 2022-03-04
More commands

oyvindberg:


bleep-core/src/main/scala/bleep/BuildPaths.scala
```diff
@@ -49,8 +33,28 @@ object BuildPaths {
     case object BSP extends Mode
   }
 
-  def fromBleepJson(cwd: Path, bleepJsonFile: Path, mode: Mode): BuildPaths =
-    fromBuildDir(cwd, bleepJsonFile.getParent, mode)
+  def find(cwd: Path, mode: Mode): Either[BuildException.BuildNotFound, BuildPaths] = {
+    // keep looking up until we find build file
+    def findBleepJson(from: Path): Option[Path] = {
+      def is(f: File): Option[File] =
+        Option(f.list()) match {
```
hamnis, line 40: instead of listing, why not just check if if known file exists?
new File(f, constants.BuildFileName).exists

is probably a bit faster

oyvindberg: Yeah, that should be faster, I'll fix it

BTW if you have any cycles to review more I'd be happy to break this up into more digestible PRs.

bleep/src/main/scala/bleep/BleepCommandRemote.scala
```diff
@@ -16,39 +17,59 @@ trait BleepCommandRemote extends BleepCommand {
...
 
-    projects.map(targetId).asJava
-  }
+  def buildTargets(buildPaths: BuildPaths, projects: Seq[model.CrossProjectName]): util.List[bsp4j.BuildTargetIdentifier] =
+    projects.map(p => buildTarget(buildPaths, p)).asJava
 
   def runWithServer(bloop: BloopServer): Unit
 
   override final def run(): Unit = {
+    val address = CompileServerConfig.load(started.userPaths) match {
```
hamnis, line 29: Mill automatically starts a server, and will allow you to run with --no-server if you really dont want it
maybe this is a sane default for us as well.

We can shut down by running `bleep compile-server stop` or by passing --no-server for instance.

PR #135, opened 2022-05-09
Ability to run commands from script (compile, test, etc)

oyvindberg:


oyvindberg:
doh, created a new one :D

PR #148, opened 2022-06-20
yaml instead of json

sideeffffect:
Continuing the conversation from https://github.com/oyvindberg/bleep/pull/7#issuecomment-1160089636

There are other interesting candidates for format for configuration besides JSON: JSON5, TOML, HOCON, ... However, YAML is the best choice, because:
 * it is the most widely used
 * has good Scala support
 * is a superset of JSON, allows for smooth transition for Bleep users (if they choose to do so)


oyvindberg:
Interesting, thanks! I'll play around with it tonight and gather some thoughts around this. I'm positive about it, but I won't commit to merging it yet.

If you have time, could you:
- Remove JSON filename support
- make same change in bleep.json as in build.sbt
- run tests so we can inspect what the new files will look like
?

sideeffffect:
native-image failing only on Windows :disappointed: 
```
Fatal error: com.oracle.svm.core.util.VMError$HostedError: should not reach here
	at com.oracle.svm.core.util.VMError.shouldNotReachHere(VMError.java:64)
	at com.oracle.svm.hosted.jdk.JNIRegistrationSupport.makeShimDLL(JNIRegistrationSupport.java:273)
```
This is the difference in logs
image

I guess the only things we can do is report it to GraalVM... the only issues I was able to google are these, but they don't seem relevant at first glance:
https://github.com/oracle/graal/issues/3393
https://github.com/oracle/graal/issues/3659

sideeffffect:
Reported bug for GraalVM native-image https://github.com/oracle/graal/issues/4659

sideeffffect:
@oyvindberg Let's suppose that the Windows native-image issue is sorted for a moment now.
Are you suggesting to drop the support for mere JSON? Isn't that going too far? Having the ability to go with JSON only could appeal to some people, YAML isn't uncontroversial. And since JSON is a subset of YAML, we can have both.

oyvindberg:
Absolutely, if we do yaml we drop json. There are no legacy users to consider, and I only want flexibility in this tool where it is absolutely needed.

sideeffffect:
Cool, fair enough :+1:

oyvindberg:
I pushed a commit now with further changes, you can also see yaml versions of the build file for bleep itself, and for the four projects which are used for testing. the test suite does a full import of an sbt build for each project and generates a corresponding bleep.yaml file so we can track changes in behavour over time

oyvindberg:
the build files do look a whole lot better, I'll give it that

hamnis:
1. fan of choosing one format
2. dont let that format be yaml, as there are [too many](https://yaml-multiline.info/) [known problems](https://www.arp242.net/yaml-config.html)  with [it](https://www.bram.us/2022/01/11/yaml-the-norway-problem/).
3. [yaml is not really a subset of json](https://john-millikin.com/json-is-not-a-yaml-subset)

oyvindberg:
I'm sorry @hamnis, it's going to be json or yaml. Someone else will have to fight the good fight of making the world choose a slightly less terrible way of expressing data.

Prioritized list of things I care about:
1) ubiquitous tooling support (syntax highlighting, json schema or similar with auto-complete in IDE)
2) familiarity
3) readability

slight win for json on 1) and 2), while yaml wins 3) by a very large margin.

Have a look at [this build file](https://github.com/oyvindberg/bleep/blob/144e3289e48f6ac4194a478a9260fae2731a2c1f/snapshot-tests/doobie/bleep.yaml) for instance. I think it looks rather nice!

I strongly dislike yaml as well for all the reasons you gave, but it's better to read and write than json. and no other data language is close in terms of tooling support OOTB and familiarity.

oyvindberg:
that said, this snakeyaml library needs to be replaced somehow. are there any other options out there?

oyvindberg:
maybe as a multi-line string instead, I've heard those are awesome in yaml :D

oyvindberg:
turns out there is a successor library called snakeyaml-engine, which only does yaml 2.4 (which seems like it avoids some of the common problems with yaml). it seems to be of much higher quality, so I don't suppose it will cause the same issue. crucially, it has no support for the old, bastardized java beans pattern at all.

I forked circe-yaml, updated it to that new library. necessary changes https://github.com/circe/circe-yaml/compare/master...bleep-build:snakeyaml-engine?expand=1 . There are tests for some old yaml-crap which need to be updated before it can be upstreamed.

oyvindberg:
haha cool, I read that as 2.4 and kept writing that everywhere

sideeffffect:
```yaml
      options:
      - -encoding
      - utf8
      - -feature
      - -language:experimental.macros
      - -language:higherKinds
      - -language:implicitConversions
      - -unchecked
```
can be just
```yaml
options: ['-encoding', 'utf8', '-feature', '-language:experimental.macros', '-language:higherKinds', '-language:implicitConversions', '-unchecked']
```

oyvindberg:
ok, we'll iterate on this going forward, but I'm feeling good enough about it now. In it goes, thanks @sideeffffect

sideeffffect:
Nice :rocket: thank you too

PR #159, opened 2022-07-04
improve scalajs support

oyvindberg:


oyvindberg:
not exactly what it says on the tin, but let's land the node management code

PR #166, opened 2022-07-14
Docs

oyvindberg:
- [x] port docusaurus and mdoc sbt plugins
- [x] generate docusaurus site
- [x] publish to `https://bleep.build`
- [ ] write docs


oyvindberg:
still a bunch of content missing, but let's say this is enough for PR number one. work continues

PR #171, opened 2022-07-19
Speedup CI build somewhat

oyvindberg:
the `ENV: CI` was lost at some point, and with it the sneaky coursier cache where the snapshot tests place resolving cache within the coursier folder. 

there is still more to be done here

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,8 @@ jobs:
         run: cs install --channel channel bleep --verbose
 
       - name: Run tests
+        env:
+          CI: true
         run: |
           bleep generate-resources
           bleep test
@@ -74,6 +76,8 @@ jobs:
         if: runner.os != 'Windows'
 
       - name: Test binary after build (non-windows)
+        env:
+          CI: true
         run: ./${{ matrix.file_name }} --ignore-version-in-build-file test --no-color jvm213
         if: runner.os != 'Windows'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -29,6 +29,8 @@ jobs:
         run: cs install --channel channel bleep --verbose
 
       - name: Run tests
+        env:
+          CI: true
```
sideeffffect, line 33: Isn't this redundant? :thinking: 
https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables=

oyvindberg: Yeah, that's probably why I deleted the declarations in the first place, heh :)

I was trying to get the first branch taken [here](https://github.com/oyvindberg/bleep/blob/master/bleep-cli-test/src/scala/bleep/IntegrationSnapshotTests.scala#L24-L26), but it somehow still didn't work. I just merged this branch without digging too much, as it'll need to be revisited very soon anyway.

The integration snapshot tests imports a bunch of real, complex sbt builds into bleep, and makes a bunch of property tests on the result. Currently all these other builds are git submodules which need to be checked out for every build, and coursier needs to resolve all the dependencies for all the builds. as a result the github action cache is absolutely enormous. My plan is to replace submodules with code which checks out the projects when needed, and try to wrestle coursier into resolving dependencies without actually downloading them.

PR #173, opened 2022-07-26
Model: Add strict mode

oyvindberg:
when true, add all scalac options for sbt-tpolecat to the given project


snapshot-tests/tapir/bleep.yaml
```diff
@@ -1283,7 +1274,13 @@ templates:
       options: -Vimplicits -Vtype-diffs -Wconf:cat=other-match-analysis:error -Wdead-code
         -Wextra-implicit -Wnumeric-widen -Wunused:explicits -Wunused:implicits -Wunused:imports
         -Wunused:locals -Wunused:nowarn -Wunused:params -Wunused:patvars -Wunused:privates
-        -Wvalue-discard -Xlint:-byname-implicit -Xlint:deprecation -Xlint:strict-unsealed-patmat
+        -Wvalue-discard -Xlint:-byname-implicit -Xlint:adapted-args -Xlint:constant
```
sideeffffect, line 1277: Why are all these new options here? Shouldn't adding `strict: true` be enough, so that all these options don't have to be stated explicitly?

oyvindberg: It is likely mismatched versions of sbt-tpolecat. If the set of flags discovered in the (in this case) tapir build do not contain *all* those produced by sbt-tpolecat for that particular scala version, the replacement with `strict: true` is not done.

PR #177, opened 2022-07-28
Common trait BleepScript for all Bleep scripts

sideeffffect:


oyvindberg:
Thanks, s bit less typing is good :)

PR #183, opened 2022-08-03
Move implementation to BleepScriptRunner

sideeffffect:
I see that you have refactored [`BleepScript` to just one class](https://github.com/oyvindberg/bleep/commit/c1b0ddb425b5d5ba6a595f798e6f4a2dc83fb9bf). Feel free to close this PR if you're happy with the current design.

But if you entertain me for a sec, I'd like to present to you the rationale for having both `BleepScript` _and_ `BleepScriptRunner`: It seemed to me a better, more future-proof design to have
 * just an interface, _unburdened by any implementation_, so that we can have a way to refer to _all_ possible plugins, e.g. `List[BleepScript]`, or `Map[String, BleepScript]`, and
 * a class to make it easier for implementers to create Bloop scripts, hence `BleepScriptRunner`

Yes, it's two classes instead of one, but the additional complexity is _very_ small, so I think it's worth having it. The benefit being more clean and future-proof design.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,8 @@ paths and so on.
 import bleep.{bootstrap, model}
 import bleep.tasks._
 
-object GenNativeImage extends App {
-  bootstrap.forScript("GenNativeImage") { started =>
+object GenNativeImage extends BleepScriptRunner("GenNativeImage") {
+  override def run(started: Started, commands: Commands, args: List[String]): Unit = {
     val projectName = model.ProjectName("myproject")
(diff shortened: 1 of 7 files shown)


oyvindberg:
I don't think I see the need. The way to talk about more than one script is `scripts: JsonMap[ScriptName, JsonList[ScriptDef]]` from the model. all the user written scripts may tap into different platforms and so on, so there is no guarantee either way that you can put them all in a list.

PR #184, opened 2022-08-04
Use BleepScript at more places

sideeffffect:


oyvindberg:
nice, thanks. good to have it consistent

PR #197, opened 2022-10-25
Direct users to documentation

sideeffffect:
Maybe it's better to direct users straight to the documentation section of the website. At least for the time being. The website https://bleep.build/ itself still has the default Docusaurus content.


oyvindberg:
thanks! It was that way as the docs as just 50% there, it takes quite a lot of time to find the motivation! :)

PR #201, opened 2022-10-31
Add `publish-local` and `dist` commands

oyvindberg:
Still beneficial to write a script for it to thread through version, groupId, wanted projects

### dist

https://user-images.githubusercontent.com/247937/199126957-22bfae82-a86a-4974-bca8-26f75b713296.mp4

### publish-local

https://user-images.githubusercontent.com/247937/199127010-d18120da-1754-4f8c-bcd4-742a003706ab.mp4


bleep-core/src/scala/bleep/packaging/PackagedLibrary.scala
```diff
@@ -0,0 +1,6 @@
+package bleep
+package packaging
+
+import coursier.core.Dependency
+
+case class PackagedLibrary(asDependency: Dependency, files: Layout[RelPath, Array[Byte]])
```
sideeffffect, line 6: Array's tend to be a pain in the neck. Would [`ArraySeq`](https://www.scala-lang.org/api/2.13.x/scala/collection/immutable/ArraySeq.html) also work here?

oyvindberg: I sympathize, but not in this case. `Array[Byte]` is an optimal type for interop between different code paths (read from disk, extract from string, read from network) and then flushing to disk through `Files.write` or similar. 

I suppose we could hide it behind a `private[bleep]` modifier inside some wrapper class, but if you're the type to mutate random byte arrays coming through your code I don't suppose a `private` modifier will stop you :D

sideeffffect: `ArraySeq` is such a wrapper around `Array`, easy to turn into and from, back and forth, conveniently included in the standard library.
But hey, this is not a big deal.

hamnis: If one would want a immutable bytearray you can use ByteVector from scodec-bits,
In this case this probably is not really that interesting.

bleep-core/src/scala/bleep/packaging/createJar.scala
```diff
@@ -0,0 +1,54 @@
...
+import bleep.{model, RelPath}
+
+import java.io.ByteArrayOutputStream
+import java.nio.file.attribute.FileTime
+import java.nio.file.{Files, Path}
+import java.util.jar.{Attributes, JarEntry, JarOutputStream, Manifest}
+import scala.collection.mutable
+
+object createJar {
+  private val epochTime = FileTime.fromMillis(0)
```
sideeffffect, line 12: epoch 0 tends to cause issues; https://github.com/sbt/sbt/pull/6254
sbt and Bazel use 2010.

So `1262304000000`, not `0`.
```suggestion
  private val epochTime = FileTime.fromMillis(1262304000000)
```

oyvindberg: nice, thanks!

PR #202, opened 2022-11-01
add `fmt` command

oyvindberg:
downloads scalafmt native build based on the requested version in `.scalafmt.conf`.

https://user-images.githubusercontent.com/247937/199134193-47abaad4-384e-4d56-b70d-347461c93837.mp4

### DIFF
--- a/bleep-model/src/scala/bleep/FetchScalafmt.scala
+++ b/bleep-model/src/scala/bleep/FetchScalafmt.scala
@@ -1,0 +1,33 @@
+package bleep
+
+import coursier.cache.{CacheLogger, FileCache}
+import coursier.jvm.JvmIndex
+import coursier.util.{Artifact, Task}
+
+import java.nio.file.Path
+import scala.concurrent.duration.Duration
+import scala.concurrent.{Await, ExecutionContext}
+
+object FetchScalafmt {
+  def apply(cacheLogger: CacheLogger, ec: ExecutionContext, version: String): Path = {
+    val filename = JvmIndex.defaultOs() match {
+      case "darwin" =>
+        "scalafmt-macos"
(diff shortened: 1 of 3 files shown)


bleep-model/src/scala/bleep/FetchScalafmt.scala
```diff
@@ -0,0 +1,33 @@
...
+
+object FetchScalafmt {
+  def apply(cacheLogger: CacheLogger, ec: ExecutionContext, version: String): Path = {
+    val filename = JvmIndex.defaultOs() match {
+      case "darwin" =>
+        "scalafmt-macos"
+      case "linux" =>
+        "scalafmt-linux-musl"
+      case other =>
+        throw new BleepException.Text(s"Sorry, no native scalafmt launcher for $other")
```
sideeffffect, line 19: Is it thinkable to run Scalafmt as JVM bytecode from Bloop (just as other Bleep scripts are) instead of downloading and executing platform dependent executable?

oyvindberg: It can absolutely be done, not sure if it's exactly in the spirit of what youre asking for.

The easiest is probably to resolve scalafmt artifacts through coursier and spin up a JVM outside of bloop. This could be done in a script or in bleep proper. This would be good to actually do in `bleep.commands.Scalafmt` as a fallback for windows users.

I guess you can also resolve and classload scalafmt inside a script as well.

oyvindberg: That said, the native launcher should have massive performance benefits. I'm personally happy to finally have a scalafmt launcher which resolves newest native :)

PR #230, opened 2022-12-19
Introduce `model.BuildVariant` to deal with different versions of the build

### LINKED ISSUE #228: scripts can bootstrap a rewritten variant of the build
rewrites are already in place, but some surface is missing before scripts can do arbitrary rewrites before bootstrapping

note: rewrites are named, so any combination of rewrites will yield a new build folder.

for instance if you bootstrap a build with `add-jmh` and `upgrade-deps` the bloop folder would end up in something like `./.bleep/builds/add-jmv__upgrade-deps/.bloop`

oyvindberg:
Enable scripts to provide rewrites to produce variants

You'll now have a .bleep directory with the given layout:
```
.bleep
  builds
    normal
    bsp
    updated-deps__enable-smh__enable-graalvm-agent
```
each of the subdirs under `builds` contain their own `.bloop` directory, logs and related files

Fixes #228


oyvindberg:
Example:
```scala

object GenDocumentation extends BleepScript("GenDocumentation") {
  override val rewrites: List[BuildRewrite] = List(
    new BuildRewrite {
      override val name = model.BuildRewriteName("add-dummy-flag")

      override protected def newExplodedProjects(oldBuild: model.Build): Map[model.CrossProjectName, model.Project] =
        oldBuild.explodedProjects.map { case (crossName, p) =>
          (crossName, p.copy(scala = p.scala.map(s => s.copy(options = s.options union model.Options.parse(List("-foo"), None)))))
        }
    }
  )
// ...
}

PR #236, opened 2022-12-22
Pin gha ubuntu version to 20.04

Baccata:
Attempts to address libc incompatibility errors experienced in CIs

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build:
     timeout-minutes: 15
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     if: "!contains(github.event.head_commit.message, 'ci skip')"
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -124,7 +124,7 @@ jobs:
 
   release:
     timeout-minutes: 15
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
sideeffffect, line 127: Wouldn't it be enough to pin it only here for the `release` job?

Baccata: I think it's better to keep a consistent version. Moreover, the release phase benefits from saved artifacts from earlier phases, see https://github.com/oyvindberg/bleep/pull/236/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R118-R123

oyvindberg: Exactly, let's keep it consistent! :)

oyvindberg:
whaat, the github action checks didn't run, and it seems this wasn't as easy as this. for one thing the old binary doesn't work, and scalafmt doesn't work. pff

oyvindberg:
merged on mobile so I didn't see that

sideeffffect:
Should we start enforcing the protection of the `master` branch?

oyvindberg:
Absolutely, doing it now. I hadn't really thought about what it meant that it was disabled :)

Issue #226, opened 2022-12-19
support shell commands in `scripts`


oyvindberg:
- do we need to string-template in any values?

PR #244, opened 2022-12-22
Split script declarations in two, and add support for running shell scripts

### LINKED ISSUE #227: build: don't bake several values into a string
`jvm/name` contains both name and version
`scripts` contain both module and class name

oyvindberg:
Fixes #227 and #226

- in the shell script structure `command` and `override-os` are optional, but bleep will error out if you try to run a script on an os where you have neither. override-os takes precedence

## example

```yaml
scripts:
  generate-docs:
    main: bleep.scripts.GenDocumentation
    project: scripts
  generate-resources:
    main: bleep.scripts.GenerateResources
    project: scripts-init
  generate-videos:
    main: bleep.scripts.GenDemoVideos
    project: scripts
  it:
    main: bleep.scripts.It
    project: scripts
  ls:
    command: ls -lah
    override-os:
      darwin: |
        cd bleep-core
        echo `pwd`
        ls -lah
      windows: dir
```

in the structure `ls` provides both `command` and `override-os` are optional, but bleep will error out if you try to run a script on an os where you have neither. `override-os` takes precedence

### DIFF
--- a/bleep-model/src/scala/bleep/model/ScriptDef.scala
+++ b/bleep-model/src/scala/bleep/model/ScriptDef.scala
@@ -1,21 +1,42 @@
 package bleep.model
 
-import io.circe.{Decoder, DecodingFailure, Encoder, Json}
+import io.circe.generic.semiauto.deriveCodec
+import io.circe._
 
-case class ScriptDef(project: CrossProjectName, main: String)
+sealed trait ScriptDef
 
 object ScriptDef {
-  implicit val decodes: Decoder[ScriptDef] = Decoder.instance(c =>
-    c.as[String].flatMap { str =>
-      str.split("/") match {
-        case Array(projectName, main) =>
-          CrossProjectName.decodes.decodeJson(Json.fromString(projectName)).map(crossProjectName => ScriptDef(crossProjectName, main))
+  case class Main(project: CrossProjectName, main: String) extends ScriptDef
+  object Main {
+    implicit val codec: Codec.AsObject[Main] = deriveCodec
+  }
 
-        case _ =>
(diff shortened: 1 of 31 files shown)


oyvindberg:
the old string syntax is still supported, but for instance `bleep build normalize` will write the scripts in this object shape instead, so it's for migration

oyvindberg:
WDYT @sideeffffect ? I think I want to keep `scripts` in the end, mostly for pedagogical reasons - it's much easier to see it and guess how it works that way

bleep-model/src/scala/bleep/model/ScriptDef.scala
```diff
@@ -1,21 +1,42 @@
...
-          CrossProjectName.decodes.decodeJson(Json.fromString(projectName)).map(crossProjectName => ScriptDef(crossProjectName, main))
-
-        case _ =>
-          Left(DecodingFailure(s"$str needs to be on the form `projectName(@crossId)/fully.qualified.Main`", c.history))
+  case class Main(project: CrossProjectName, main: String) extends ScriptDef
+  object Main {
+    implicit val codec: Codec.AsObject[Main] = deriveCodec
+  }
+
+  case class Shell(command: Option[String], `override-os`: Option[Map[String, String]]) extends ScriptDef
```
sideeffffect, line 14: What's the harm in making `command` mandatory?

sideeffffect (approved):
Looks fine, feel free to disregard my question :wink:

bleep-model/src/scala/bleep/model/ScriptDef.scala
oyvindberg: maybe there is value in writing a script which only works on one OS. not really sure. we can always constrain it later

Issue #235, opened 2022-12-22
Native-image assumption is too aggressive 


Baccata:
Alright, this is an interesting one : I'm attempting to run bleep at $work in a self-hosted GHA. I don't have control over the runners as the infra is provided by a completely isolated team. 

* If I install bleep using the coursier channel, as you would expect, I essentially bump into [this](https://github.com/actions/runner-images/issues/6647) : the LIBC available on the runner is not matching the one that bleep was built against. 
* So I'm like "well, this is CI anyway, let's try and run bleep on JVM, shall we, it's not the end of the world right now?" Unfortunately, when checking for the version number in the build-file (in order to download the correct version), bleep presumes it's running from a native-image, and invokes a [coursier function](https://github.com/coursier/coursier/blob/c7f9658419e1fce68e5e0c222523ee8149175bd4/modules/jvm/src/main/java/coursier/jvm/Execve.java#L10) that is unavailable on JVM (only implemented for graal). 

The workaround I found is to use coursier to bootstrap the version of bleep referenced in the build file, something like : 

``` 
BLEEP_VERSION=$(cat bleep.yaml | grep '$version' | cut -d ":" -f2 |   tr -d ' ')
cs bootstrap build.bleep:bleep-cli_2.13:$BLEEP_VERSION -M bleep.Main -o bleep
```

oyvindberg:
Hey there, thanks for testing it! I bet there are many small things like this lurking around.

Do you have any insight into this? Would it be enough if we set the GitHub runner to use an old Ubuntu runner?

sideeffffect:
It would be still nice to be able to run Bleep just on JVM just from Maven Central, right?

keynmol:
> Would it be enough if we set the GitHub runner to use an old Ubuntu runner?

That's how it's usually solved, yes. 
A lot of things are very slow moving so 22.04 is waaay too young to build any tools.
Plus you barely get any benefits by using it.

oyvindberg:
Running bleep from the JVM should absolutely work! Can you try running from the JVM again if you specify `--dev` @Baccata ? That's the flag to disable the launcher functionality. I bet some logic there is missing an `if (isNativeImage) ...`

> A lot of things are very slow moving so 22.04 is waaay too young to build any tools.
> Plus you barely get any benefits by using it.

I hadn't even considered there was a tradeoff there, just picked latest :)

[...]

PR #245, opened 2022-12-22
Build native image for ubuntu on older version

oyvindberg:
build native-image for ubuntu on older version. Fixes (#235)

Followup to #236

- I suppose it'll be forward compatible
- note: there is a migration step here. The cs launch things are to launch bleep on the JVM in order to produce a release with lower version of glibc

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   build:
     timeout-minutes: 15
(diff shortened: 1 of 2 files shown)


sideeffffect:
For similar reasons, shouldn't we be using older OSes than `macos-latest` and `windows-latest`? :thinking:

oyvindberg:
no idea! I'm tempted to say let's go with `latest` until there are problems. the migration step will be the same as in this PR - so not much work to fix it later

Issue #19, opened 2021-11-18
idea: allow projects depend on scripts. 


oyvindberg:
This will allow us to do source generation and other shenanigans.

It should be relatively easy to implement actually, some tree traversal in the `bloop compile` command to find the commands, and then tell bloop to compile the necessary steps in turn

oyvindberg:
So since I write this issue originally Bloop has gained the power of running source generators, see [this structure](https://github.com/scalacenter/bloop-config/blob/main/config/src/bloop/config/Config.scala#L260-L264).
I have not had the time to personally experiment with it.

If you can find a way to integrate with that it would be very cool indeed!

We'll need some kind of hint in the build, maybe something like this:
```yaml
projects:
  a:
    source-generator-scripts: gen-a-sources
scripts:
  gen-a-sources: ...
```

PR #247, opened 2022-12-24
Add support for code-generators

Baccata:
This test case demonstrates pretty well what it can do:

```scala
  test("resource generator") {
    val bleepYaml = """
projects:
  a:
    extends: common
    platform:
      mainClass: test.Main
    sourcegen: scripts/testscripts.SourceGen
  scripts:
    extends: common
    dependencies: build.bleep::bleep-core:${BLEEP_VERSION}
templates:
  common:
    platform:
      name: jvm
    scala:
      version: 3.2.1
"""

    val Main = """
package test

object Main {
  def main(args: Array[String]): Unit =
    println("result: " + testgenerated.GeneratedSource.result)
}
"""

    val SourceGen = """
package testscripts

import bleep.*
import java.nio.file.Files

object SourceGen extends BleepCodegenScript("SourceGen") {
  def run(started: Started, commands: Commands, crossName: model.CrossProjectName, args: List[String]): Unit = {
    val targetFile = started.projectPaths(crossName).sourcesDirs.generated / "sourcegen" / "testgenerated" / "GeneratedSource.scala"
    Files.createDirectories(targetFile.getParent)
    Files.writeString(
      targetFile,
      "package testgenerated;\n" +
      "\n" +
      "object GeneratedSource {" +
      "  val result = 100" +
      "}"
    )
    started.logger.withContext(targetFile).warn("Wrote ")
  }
}
"""

    runTest(
      "resource generator spike",
      bleepYaml,
      Map(
        RelPath.force("./a/src/scala/test/Main.scala") -> Main,
        RelPath.force("./scripts/src/scala/testscripts/SourceGen.scala") -> SourceGen
      )
    ) { (_, commands, storingLogger) =>
      commands.run(model.CrossProjectName(model.ProjectName("a"), None))
      assert(storingLogger.underlying.exists(_.message.plainText == "result: 100"))
    }
  }

```

Closes https://github.com/oyvindberg/bleep/issues/19

#### TODO 

- [x] Add tests (I'd need some guidance)

- [x] Amend json schema


bleep-core/src/scala/bleep/GenBloopFiles.scala
```diff
@@ -328,6 +328,27 @@ object GenBloopFiles {
         case _                                            => bloop.config.Tag.Library
       }
 
+    val sourceGenerators = {
+      val sg = explodedProject.`source-generator-scripts`.values.map { sc =>
+        if (!build.scripts.contains(sc.scriptName)) {
+          sys.error(s"Unknown script ${sc.scriptName} referenced by ${crossName.value}")
+        }
+        val codegenPath = buildPaths.generatedSourcesDir(crossName)
```
oyvindberg, line 336: let's drop the arguments thing in this first version. if we do it (and I suspect we won't have to) we'll do it across the board in a later PR.

bleep-core/src/scala/bleep/GenBloopFiles.scala
```diff
@@ -328,6 +328,27 @@ object GenBloopFiles {
...
+        val scriptCall = List("bleep", sc.scriptName.value)
```
oyvindberg, line 337: to increase the chance of this working, we should grab the absolute path to `bleep` used to run this. You can see that done in `SetupIde#run()`. Try to extract that into it's own thing and reference it both places.

bleep-core/src/scala/bleep/GenBloopFiles.scala
```diff
@@ -328,6 +328,27 @@ object GenBloopFiles {
...
+        val `this` = "$" + "this"
+        // Dereferencing all '$this' into the cross name, so that users can
+        // inject the module's name into the script.
+        val args = sc.arguments.map {
+          case `this` => crossName.value
+          case other  => other
+        }
+        val fullCall = scriptCall ++ args
+        // TODO what are the globs for ?
```
oyvindberg, line 346: The source globs are likely used for cache invalidation - any sources changed and the resource generator is run again.

I guess that if we pull all the transitive source directories of the script project that should work, something like this:
```scala
    val scriptSources: List[Path] =
      started.build
        .scripts(scriptName)
        .values
        .flatMap {
          case model.ScriptDef.Main(scriptProjectName, _) =>
            val withTransitive = List(scriptProjectName) ++ started.build.transitiveDependenciesFor(scriptProjectName).keys
            withTransitive.flatMap { projectName =>
              val bloopProject = started.bloopProjects(projectName)
              bloopProject.sources ++ bloopProject.resources.getOrElse(Nil)
            }
          case model.ScriptDef.Shell(_, _) =>
            throw new BleepException.Text(s"cannot use shell script as resource generator, because bleep doesn't know how to invalidate it")
        }
        .distinct

```

oyvindberg (commented):
Thanks, I'm eager to check it out! :)

bleep-core/src/scala/bleep/GenBloopFiles.scala
oyvindberg: As you see I think for shell scripts the user will need to provide this invalidation information themselves

bleep-core/src/scala/bleep/GenBloopFiles.scala
Baccata: I'd rather not drop it, at least not without a way to communicate to the script what module it was triggered for (unless there's a way to achieve it without args, which I doubt there is). 

What I have codified in this draft is that "$this" passed in an arg list would be a way to do just that.

bleep-core/src/scala/bleep/GenBloopFiles.scala
Baccata: Cool, I wanted to ask about that 👍

bleep-core/src/scala/bleep/GenBloopFiles.scala
Baccata: Yeah the user might want to provide invalidation information for scripts too, if scripts depend on reading files from some arbitrary folder from the file system

bleep-core/src/scala/bleep/GenBloopFiles.scala
oyvindberg: Is the reason why you want this that you're thinking of cross projects where you need to generate the same resources many times?

bleep-core/src/scala/bleep/GenBloopFiles.scala
oyvindberg: Yeah, we should definitely add that as an optional element in `ScriptDef.Shell`. So if you want to use that shell script as a resource generator you need to provide a list of input files. probably a list of paths relative to the build directory - that is the root of the project

bleep-core/src/scala/bleep/GenBloopFiles.scala
oyvindberg: Ok, I've been thinking ten minutes. Let's keep the functionality of passing project name as parameter.

Just to explain my somewhat negative position:
- passing arguments feels like programming in YAML, which is not a favourite pastime of mine.
- I'd rather that we make each script each own entry in `scripts` and a corresponding scala `object`. super clear and obvious, easy to run, easy to debug.

However, since the bloop resource generators are connected to a bloop project, we really lose the opportunity to run/track the resource generator for all `CrossId`s in one go. That we'll just have to live with.

How about we just establish the convention that all resource generators are passed the project name as a parameter? No configuration or flexibility, just that?

bleep-cli/src/scala/bleep/sbtimport/buildFromBloopFiles.scala
```diff
@@ -161,7 +161,8 @@ object buildFromBloopFiles {
         `source-layout` = Some(sources.sourceLayout),
         `sbt-scope` = Some(projectType.sbtScope),
         isTestProject = if (projectType.testLike) Some(true) else None,
-        testFrameworks = testFrameworks
+        testFrameworks = testFrameworks,
+        `source-generator-scripts` = model.JsonList.empty // TODO can we do better ?
```
oyvindberg, line 165: Absolutely! 

Let me explain what is in place today:

after importing the full build from sbt, bleep looks at whatever generated (re-)sources are present.

this is done in [generateBuild](https://github.com/oyvindberg/bleep/blob/master/bleep-cli/src/scala/bleep/sbtimport/generateBuild.scala#L49).

If there are any, it generates a `GenerateResources` script, which can re-generate all those files. 

Obviously this approach is not-clever, since it just writes strings back to disk.

It is the intention that the user replace the implementation of the script with calls to code which can re-generate the same things. if you have generated protobuf you'll call protobuf yourself. My hope is that we can port a rather small set of sbt plugins to enable most usage of generated resources in sbt projects in bleep. Potentially we could extract for instance protobuf configuration from the sbt build, if we wanted to invest time in that.


A path forward here is that we take what is in place today, group these files by project, and generate one resource-generating script per project. Possibly with some extra logic for cross projects to make the number of generated scripts lower by using the projectName parameter discussed in the other thread.

If the rest of the PR ends up being a lot of work we can absolutely do this in a followup PR.

bleep-core/src/scala/bleep/GenBloopFiles.scala
Baccata: Well an example of why I want that is: how do you determine the target folder if you don't know the module which invokes the code generator ? Or, what if what you want to generate is dependant on the scala version of the module in question ?

bleep-core/src/scala/bleep/GenBloopFiles.scala
Baccata: Sorry, my point was that `ScriptDef.Main` should hold that information too, because inputs to code-generators might not be in the bloop project's sources, nor it's resources.

oyvindberg: That's a constraint we can decide though, and i think we should

bleep-cli/src/scala/bleep/sbtimport/buildFromBloopFiles.scala
Baccata: > If the rest of the PR ends up being a lot of work we can absolutely do this in a followup PR. 

Right, I think I'll leave that aside as I'm not a fan of the current approach : it gives an initial impression that bleep does something really smart / magic.

oyvindberg: What i described above is a migration aid when coming from sbt to get a working build immediately.

That seems magical to you?

bleep-core/src/scala/bleep/GenBloopFiles.scala
Baccata: > How about we just establish the convention that all resource generators are passed the project name as a parameter? 

Would that mean having a `GeneratingScript` class inherting the `Script` class ? That might work (from a DX perspective, that is)

> passing arguments feels like programming in YAML, which is not a favourite pastime of mine.

Yeah I don't disagree there. 

> I'd rather that we make each script each own entry in scripts and a corresponding scala object. super clear and obvious, easy to run, easy to debug.

> No configuration or flexibility, just that?

This is a little too strict for my taste, I do think it'd be interesting to allow for users to pass some configuration to the scripts in the `bleep.yaml`. Say someone wants to publish pre-built scripts as literal "plugins" : this would allow people to customise the behaviour of the plugin without having to touch 2 files (the build file to pull the dependency and configure a script project + a scala file). 

To elaborate : if you have a `metadata` field in the projects configuration that'd receive an arbitrary Json object, bleep could pass that object to scripts and script implementors could use that to retrieve configuration elements however they please. If users would rather duplicate some Scala files instead of using yaml configuration, they'd totally be able to do that. What's more : you could deep-merge the `metadata` value of all the templates (and the one of the project itself) before passing it to the script. Imho, this would be coherent with the way bleep is configured (declaratively), generally speaking, and would prevent having to compile scala code when reconfiguring some aspects of the build (which is a one of the appeals of bleep, as having to compile a Scala-written build is detrimental to the snappiness of the build's commands)./

bleep-cli/src/scala/bleep/sbtimport/buildFromBloopFiles.scala
Baccata: It is, in the sense that without looking at the sources of the generated scripts, I'll get a false sense that the script will copy the behaviour of SBT's or mill's code-generators. 

When I say "I", I'm projecting myself as a random user. I personally have a pretty good understanding of build tools for having contributed to mill (in a non trivial manner), so by curiosity I did open the generated scripts to understand how it was working. A random user might just think "oh, cool, compilation works so my protobuf codegen logic has been ported over, neat", and be in for a puzzlement or a disappointment when they discover it's not actually copying the logic but rather the outputs.

bleep-core/src/scala/bleep/GenBloopFiles.scala
Baccata: > That's a constraint we can decide though, and i think we should 

Too constraining for my taste, tbh. Like, say my build has a source-dependency (by means of git submodules) to pull shared specifications from another repo. I'd rather not clone that entire repo in the resources of my project, so I'd clone it at the root of my project for instance, and I'd configure my code-generator to understand where the clone is located. But then, since bloop is responsible for deciding when to call my code-generator, I'd need to tell bloop to monitor the clone for changes (for when I run `git submodule update -remote`)

Baccata: Or another example closer to my heart : in the [smithy4s](https://github.com/disneystreaming/smithy4s) project I maintain, we have a [sample specs](https://github.com/disneystreaming/smithy4s/tree/series/0.17/sampleSpecs) folder that contains smithy specifications that are used both in code-generation for test-purposes, and in documentation (which is an mdoc project). Forcing a pattern where codegen inputs would HAVE to be resources of a script project would mean that it'd be harder to share the specs between my tests and my documentation. 

I also like the fact that this folder is at the root, because of how it makes discoverability of examples easier when checking-out the project

bleep-core/src/scala/bleep/GenBloopFiles.scala
oyvindberg: > Would that mean having a GeneratingScript class inherting the Script class ? That might work (from a DX perspective, that is)
Yes - that seems fitting! :)

> This is a little too strict for my taste, I do think it'd be interesting to allow for users to pass some configuration to the scripts in the bleep.yaml. Say someone wants to publish pre-built scripts as literal "plugins" : this would allow people to customise the behaviour of the plugin without having to touch 2 files (the build file to pull the dependency and configure a script project + a scala file).

My immediate thought is that bleep has not one but three main goals:
1) as simple as possible
2) as fast as possible
3) automatic import from sbt

What you're suggesting here works towards 2), but against 1). I actually think 1) may be more important, and "one way to do it" is hugely important for 1). That way is to write scala scripts.

Anyways, the merging of metadata fields appeals to me. Give me a day or two to chew on this :)

bleep-core/src/scala/bleep/GenBloopFiles.scala
oyvindberg: You can add additional resource folders! :)
```yaml
projects:
  a:
    resources: ../sampleSpecs
```

How about that?

bleep-cli/src/scala/bleep/sbtimport/buildFromBloopFiles.scala
oyvindberg: Yeah, makes sense.

In an attempt to communicate this the build import will print [this warning](https://github.com/oyvindberg/bleep/blob/master/bleep-cli/src/scala/bleep/sbtimport/generateBuild.scala#L87) in red after importing, and the script itself will print [this warning](https://github.com/oyvindberg/bleep/blob/master/snapshot-tests/bloop/imported/scripts/src/scala/scripts/GenerateResources.scala#L10) when it runs.

oyvindberg: And like i said elsewhere, we could extend the import to recognize common source generation plugins like `sbt-boilerplate` and actually set it up properly :)

Baccata: Yup that's fair, hadn't noticed the warning (or had forgotten about it), I've not tried to import an sbt project in a while. 

TBH my opinion on this should be taken with a grain of salt (I 100% respect the fact that OSS authors are entitled to driving the design however they want), and I'm not even saying the feature shouldn't exist, but rather that the feature doesn't interest me and therefore my motivation to work on that part is nil 😅

bleep-core/src/scala/bleep/GenBloopFiles.scala
Baccata: mmm okay yeah, I guess

bleep-cli/src/scala/bleep/sbtimport/buildFromBloopFiles.scala
oyvindberg: Haha fine, don't worry about it :)

For me it's crucial that there is a a one-command import which produces a working build (goal 3) so new users can test bleep in real conditions. Only after trying it out do you have to invest time into resource generation scripts.

bleep-core/src/scala/bleep/GenBloopFiles.scala
Baccata: Right, so I've thought about this some more as I've resumed implementing, and I think some configuration will be required, if we want to delegate the invocation of the codegeneration script to bloop. Taking the example of proto codegenerators, I think it's pretty natural for the proto files to be located in the folder of the project that would receive the generated code. 

So what bloop configuration should be generated when that's the case ? 

* the script source-path (transitive sources/resources) doesn't suffice because that doesn't encompass the target project's folder 
* `project-folder/**` would be overkill because that'd encompass the Scala sources and you probably don't want to re-trigger code-generation whenever an unmanaged source directory gets touched, that'd be wildly inefficient
* the resources of the project would be too prescriptive, because people may not want to package the input of the code-generators they run in end artifacts  
* you can't solve that problem in the script implementation, because bloop ultimately decides whether to run stuff or not, and therefore it may skip the codegenerators in situations where you'd want for it to run  

So my conclusion is that you have to let the user have control over the globs. 

The alternative would be to not rely on bloop for this, and have bleep invoke the code-generators itself. It'd probably make for a better user experience in the end, because you could then solve the caching within the script implementation itself, in a way that'd be convenient for the user.

oyvindberg: > I think it's pretty natural for the proto files to be located in the folder of the project that would receive the generated code.

Agreed, this will be the case in ~all existing projects.

> The alternative would be to not rely on bloop for this, and have bleep invoke the code-generators itself. It'd probably make for a better user experience in the end, because you could then solve the caching within the script implementation itself, in a way that'd be convenient for the user.

This is what I originally had in mind, but let's keep this as a last resort? I think from a performance perspective it's best to let bloop handle it, otherwise we'll need to ask bloop to compile in multiple BSP requests.

I agree with your reasoning. If input files for codegen live inside a folder which belongs to a project instead of the script project, the user will have to provide invalidation globs for it. There is no other way in that case.

How about this then? The default invalidation globs are that of the script project (sources/resources). We want users to put them there, because there is less configuration. The user is allowed to add additional globs to the script definition if the codegen script reads files outside the script project. Is that pretty much all we need?

Baccata: > How about this then? The default invalidation globs are that of the script project (sources/resources). We want users to put them there, because there is less configuration. The user is allowed to add additional globs to the script definition if the codegen script reads files outside the script project. Is that pretty much all we need?

I think that sounds about right for this first iteration 👍

Baccata:
@oyvindberg I'm very confused as to why the build is failing due to unused import. I seem to be able to reproduce locally, but the import is very much used. Possibly a compiler bug ?

oyvindberg:
That's a warning, the error is further up, involving LazyList. You're probably seeing some complexity with 2.12 cross build and collections compat

oyvindberg:
For tests I think I'll point you towards https://github.com/oyvindberg/bleep/blob/e2b09726ea93379388382c758c59976b31a1b6fe/bleep-tests/src/scala/bleep/IntegrationTests.scala#L57

I just added that last night, so it might not be optimal yet

I hope to have some time for tonight or tomorrow to look at the code in the PR :)

oyvindberg:
Thinking about the tests a bit further, that test harness is not powerful enough as is. I think it will work fine for shell script resource generators, but not for JVM source generators if we want to pull in locally built bleep code.

So we will need some tricks to talk bloop into running scripts with locally built bleep code. 
Either:
1) something similar to sbt-scripted (publish-local bleep, thread through version number to tests, native-image). This screams slow turnaround time
2) or write do some magic in  `GenBloopFiles` so bloop is instructed to start a script in a JVM with bleep class directories.

How about you focus on writing a test or two for shell resource generators using that harness I gave you first, and I'll see if I can get 2) implemented in the mean time?

bleep-core/src/scala/bleep/GenBloopFiles.scala
```diff
@@ -328,6 +331,52 @@ object GenBloopFiles {
...
+            .values
+            .flatMap {
+              case model.ScriptDef.Main(scriptProjectName, _, _) =>
+                val withTransitive = List(scriptProjectName) ++ build.transitiveDependenciesFor(scriptProjectName).keys
+                withTransitive.flatMap { projectName =>
+                  val bloopProject = getBloopProject(projectName).project
+                  bloopProject.sources ++ bloopProject.resources.getOrElse(Nil)
+                }
+              case model.ScriptDef.Shell(_, _) =>
+                throw new BleepException.Text(s"cannot use shell script as resource generator, because bleep doesn't know how to invalidate it")
```
oyvindberg, line 358: we can use shell scripts now if we add `sourceGlobs` to `ScriptDef.Shell` and assert here that it's non-empty

bleep-core/src/scala/bleep/GenBloopFiles.scala
```diff
@@ -328,6 +331,52 @@ object GenBloopFiles {
         case _                                            => bloop.config.Tag.Library
       }
 
+    val sourceGenerators = {
+      val sg = explodedProject.`source-generator-scripts`.values.map { scriptName =>
+        if (!build.scripts.contains(scriptName)) {
```
oyvindberg, line 336: style:

```scala
val script = build.scripts.get(scriptName).getOrElse(
  throw new BleepException.Text(crossName, s"Unknown script $scriptName")
)
```

Avoid looking up `build.scripts.get(scriptName)` further down.

Difference between `sys.error` and `BleepException` in bleep is that the latter will not show a stack trace to the user (only under `--debug` or in the log file)

bleep-core/src/scala/bleep/internal/ScriptShelloutCommand.scala
```diff
@@ -27,7 +27,7 @@ object ScriptShelloutCommand {
   def getForScript(scriptDef: ScriptDef, bleepExecutable: BleepExecutable): List[String] = scriptDef match {
     case Main(project, main, _) =>
       val bleepPath = bleepExecutable.getPath().toString
-      List(bleepPath, "run", "--main", main, project.name.value)
+      List(bleepPath, "run", "--class", main, project.name.value)
```
Baccata, line 30: it'd be great if this was aligned with `scala-cli`'s   `-M, --main-class`

oyvindberg: Agreed! Feel free to change it everywhere, now or in a follow-up

oyvindberg:
Ok, so I finally had time to sit down and write that test and get a feel for the code. Details for what was necessary to get the test running in commit message if that is relevant. Anyway, let's have a look at what a small example now looks like!

```scala
  test("resource generator") {
    val bleepYaml = """
projects:
  a:
    extends: common
    platform:
      mainClass: test.Main
    source-generator-scripts: sourcegen
  scripts:
    extends: common
    dependencies: build.bleep::bleep-core:${BLEEP_VERSION}
scripts:
  sourcegen: scripts/testscripts.SourceGen
templates:
  common:
    platform:
      name: jvm
    scala:
      version: 3.2.1
"""

    val Main = """
package test

object Main {
  def main(args: Array[String]): Unit =
    println("result: " + testgenerated.GeneratedSource.result)
}
"""

    val SourceGen = """
package testscripts

import bleep.*
import java.nio.file.Files

object SourceGen extends BleepCodegenScript("SourceGen") {
  def run(started: Started, commands: Commands, crossName: model.CrossProjectName, args: List[String]): Unit = {
    val targetFile = started.projectPaths(crossName).sourcesDirs.generated / "sourcegen" / "testgenerated" / "GeneratedSource.scala"
    Files.createDirectories(targetFile.getParent)
    Files.writeString(
      targetFile,
      "package testgenerated;\n" +
      "\n" +
      "object GeneratedSource {" +
      "  val result = 100" +
      "}"
    )
    started.logger.withContext(targetFile).warn("Wrote ")
  }
}
"""

    runTest(
      "resource generator spike",
      bleepYaml,
      Map(
        RelPath.force("./a/src/scala/test/Main.scala") -> Main,
        RelPath.force("./scripts/src/scala/testscripts/SourceGen.scala") -> SourceGen
      )
    ) { (_, commands, storingLogger) =>
      // need to compile `scripts` project manually since for test we run the resource generator directly with jvm
      commands.compile(List(model.CrossProjectName(model.ProjectName("scripts"), None)))

      commands.run(model.CrossProjectName(model.ProjectName("a"), None))
      assert(storingLogger.underlying.exists(_.message.plainText == "result: 100"))
    }
  }

```

So that's awesome, thank you for the work so far!

I'm wondering if it's worth doing this (now or later):
```diff
@@ -126,12 +126,10 @@ projects:
     extends: common
     platform:
       mainClass: test.Main
-    source-generator-scripts: sourcegen
+    source-generators: scripts/testscripts.SourceGen
   scripts:
     extends: common
     dependencies: build.bleep::bleep-core:${BLEEP_VERSION}
-scripts:
-  sourcegen: scripts/testscripts.SourceGen
 templates:
   common:
     platform:
```
that way there will be a bit less string matching, but it will be more complicated to run yourself. thoughts?

bleep-core/src/scala/bleep/GenBloopFiles.scala
oyvindberg: I lifted this conversation to https://github.com/oyvindberg/bleep/issues/272 and will close the thread

bleep-cli/src/scala/bleep/sbtimport/buildFromBloopFiles.scala
oyvindberg: I'm resolving the conversations, just leaving this open as a todo my myself after merge

Baccata:
> So that's awesome, thank you for the work so far!

My pleasure, and sorry I couldn't find more time to attempt in myself. Was planning on resuming over the weekend. 

> I'm wondering if it's worth doing this (now or later):
> that way there will be a bit less string matching, but it will be more complicated to run yourself. thoughts?

I think it makes total sense and it should be done in this PR, it's much more intuitive and coherent as code-generation is always gonna be tied to a project (not necessarily exploded, though). A`bleep codegen project` command could be added to help users run code-generation in isolation.

oyvindberg:
> A bleep codegen project command could be added to help users run code-generation in isolation.

Yeah, great idea - that's exactly what's missing! :)

bleep-model/src/scala/bleep/model/ScriptDef.scala
```diff
@@ -1,24 +1,24 @@
...
 
-import io.circe.generic.semiauto.deriveCodec
+import bleep.RelPath
 import io.circe._
-import io.circe.syntax.EncoderOps
+import io.circe.generic.semiauto.deriveCodec
 
 sealed trait ScriptDef {
-  def sourceGlobs: Option[SourcesGlobs]
+  lazy val asJson: Json = ScriptDef.encodes(this)
```
Baccata, line 8: out of curiosity, any reason why you don't like the `def sourceGlobs` at the sealed-trait level ?

oyvindberg: Normally there is absolutely nothing wrong with that. In this particular case, as of right now, the two `sourceGlobs` do *not* mean the same thing.

```
            script match {
              case model.ScriptDef.Main(project, _, sourceGlobs) =>
                val projectDir = buildPaths.project(project, build.explodedProjects(project)).dir
                sourceGlobs.values.toList.map(relPath => Config.SourcesGlobs(projectDir / relPath, None, Nil, Nil))
              case model.ScriptDef.Shell(_, _, sourceGlobs) =>
                val projectDir = buildPaths.project(crossName, explodedProject).dir
                sourceGlobs.values.toList.map(relPath => Config.SourcesGlobs(projectDir / relPath, None, Nil, Nil))
            }
          }
```

If you look carefully, the relative paths are added to two different absolute paths, one is relative to the script project and one relative to the project which receives the sources. This is what I talked about in one of the comments above. So no usage of the shared `sourceGlobs` would be correct. That's why I removed it

Baccata: Oh, gotcha 👍

oyvindberg:
Alright, I had some more time here @Baccata . Kindly forgive the force push, I rebased and squashed everything on top of `master`. Notably #276 went in, with considerable fallout in this branch so I didn't bother keeping the full history.

I also implemented the `bleep sourcegen` command, so most things are in place now. 
my todo list:
- implement sourcegen for sbt import
- update json schema
- testing

I did some brief testing last night. Found a way to cause eternal loops with sourcegen -> compile -> sourcegen, and was a bit bothered by sourcegen not being invalidated properly. `shell` scripts need some more love, since it's difficult to figure out what is happening for the user. Shell scripts could potentially be disabled for now

bleep-core/src/scala/bleep/GenBloopFiles.scala
```diff
@@ -11,73 +10,173 @@ import io.github.davidgregory084.{DevMode, TpolecatPlugin}
...
+        translateProject(pre, resolver, crossName, project, build, getBloopProject = get, genScriptRunner)
       }
-      SortedMap.empty[model.CrossProjectName, Lazy[Config.File]] ++ map
-    } else {
-      logger.warn(s"Refreshing ${buildPaths.bleepBloopDir}...")
+  }
 
-      val bloopFiles =
-        InMemory(logger, buildPaths, lazyResolver, build, fetchNode)
+  /** For integration tests we want to write builds which depend on the version of bleep you're working on, and we want to use the normal syntax (dependency
```
oyvindberg, line 66: move to test scope, along with other test harness related things

bleep-core/src/scala/bleep/internal/ScriptShelloutCommand.scala
```diff
@@ -0,0 +1,18 @@
...
+    val bareCommand: String = scriptDef.`override-os`.flatMap(_.get(OsArch.current.os)).orElse(scriptDef.command).getOrElse {
+      throw new BleepException.Text(s"no command found for os ${OsArch.current.os}")
+    }
+    OsArch.current.os match {
+      case bleep.model.Os.Windows =>
+        List("cmd.exe", "/C", bareCommand)
+      case _ =>
+        val quote = '"'.toString
+        val quotedQuote = "\\\\"
+        List("bash", "-c", bareCommand.replace(quote, quotedQuote))
```
oyvindberg, line 15: not sure about this replacement. notable the `quotedQuote` is missing a quote. maybe just write script to temporary file instead?

Baccata: Tbh this was copied verbatim from [there](https://github.com/oyvindberg/bleep/pull/247/files/d138225085c2e4973711e1ef0b614802e15dff28#diff-0159507d8e14da3d810ed850041a4895e25f75d74cb4bec8fba01265fb904a8bL33-L39), so I can't comment about why the replacement is there :/

Baccata:
>  Found a way to cause eternal loops with sourcegen -> compile -> sourcegen 

out of curiosity, what was the way in question ?

oyvindberg:
> > Found a way to cause eternal loops with sourcegen -> compile -> sourcegen
> 
> out of curiosity, what was the way in question ?

this, it set up invalidation glob on the generated sources folder :)

https://github.com/oyvindberg/bleep/pull/247/files#diff-e314efdcf1d13d7031d025d7beaa46ba2d8160e994a1c7f3fbcbc687a8a7a450R455

bleep-core/src/scala/bleep/internal/ScriptShelloutCommand.scala
oyvindberg: Right. Also no idea :D gonna have to test it then

bleep-core/src/scala/bleep/BleepExecutable.scala
```diff
@@ -0,0 +1,103 @@
...
+        case n  => Some(n)
+      }
+  }
+
+  def findCurrentBleep(logger: Logger): Option[BleepExecutable] = {
+    def complain(msg: String): None.type = {
+      logger.warn(s"Problem while discovering how to run the bleep you started: $msg. Falling back to downloading latest release.")
+      None
+    }
+    val currentInfo = ProcessHandle.current.info()
```
Baccata, line 69: 🎉

oyvindberg: This was a really great find! No idea why all those Google searches i did in the past didn't bring it up.

I have some time for bleep today, i think I'll merge this in one form or another.

The experience has some rough edges, and i feel like the whole thing is exactly on the edge of the complexity budget of what should go in, but we need this somehow. 

The integration tests in particular are difficult to understand and brittle.

Also things like how bloop behaves when the sourcegen is incorrectly setup is not optimal. Invalidation is also not on point, bloop does not always pick up changes to sourcegen scripts.

Baccata: Does this imply that it'd be preferable longer term for bleep to own the code-generation logic ?

oyvindberg: Maybe! I think it's best that the logic for running these things live in bloop, because it can know better when to run them. I think I've been observing bugs in the implementation there, and those can be reported and fixed.

Baccata: 👍

oyvindberg:
I'll go ahead and merge this shortly.

I don't believe it's perfect yet, but let's not let perfect be the enemy of good. We'll improve it going forward 🎆

Baccata:
🎉

PR #251, opened 2022-12-29
Iterate on project selection

oyvindberg:
## Can now choose projects by:
- `@2` or `@3` for epoch scala versions
- `@213` for scala binary versions
- `@native`, `@js`, `@jvm` for platform
- `main`/`test` to account for `isTestProject`
- `@jvm213` still works like before

## Can also subtract projects from selection by prefixing `~` `@2 ~@jvm` will choose all projects for for instance 2.12 and 2.13 except those for jvm platform

### DIFF
--- a/bleep-model/src/scala/bleep/rewrites/keepSelectedProjects.scala
+++ b/bleep-model/src/scala/bleep/rewrites/keepSelectedProjects.scala
@@ -1,10 +1,10 @@
 package bleep
 package rewrites
 
-case class keepSelectedProjects(selectedProjectGlobs: List[String]) extends BuildRewrite {
+case class keepSelectedProjects(selectedProjectGlobs: Array[model.ProjectGlob]) extends BuildRewrite {
   override val name = model.BuildRewriteName("keep-selected-projects")
 
-  def selectedPlusTransitiveDeps(selectedProjectNames: List[model.CrossProjectName], build: model.Build): Set[model.CrossProjectName] = {
+  def selectedPlusTransitiveDeps(selectedProjectNames: Array[model.CrossProjectName], build: model.Build): Set[model.CrossProjectName] = {
     val b = Set.newBuilder[model.CrossProjectName]
     selectedProjectNames.foreach { name =>
       b += name
@@ -14,7 +14,7 @@ case class keepSelectedProjects(selectedProjectGlobs: List[String]) extends Buil
   }
   protected def newExplodedProjects(oldBuild: model.Build): Map[model.CrossProjectName, model.Project] = {
     val globs = new model.ProjectGlobs(None, oldBuild.explodedProjects)
-    val selectedProjectNames = selectedProjectGlobs.flatMap(str => globs.projectNameMap.getOrElse(str, Array.empty[model.CrossProjectName]))
+    val selectedProjectNames = selectedProjectGlobs.flatMap(globs.projectNameMap.get).selection
     val withTransitive = selectedPlusTransitiveDeps(selectedProjectNames, oldBuild)
 
     val chosen = oldBuild.explodedProjects.filter { case (name, _) => withTransitive(name) }
--- a/bleep-cli/src/scala/bleep/Main.scala
+++ b/bleep-cli/src/scala/bleep/Main.scala
@@ -3,7 +3,7 @@ package bleep
 import bleep.bsp.BspImpl
 import bleep.internal.{fatal, Os}
 import bleep.logging._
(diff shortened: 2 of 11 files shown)


bleep-model/src/scala/bleep/rewrites/keepSelectedProjects.scala
```diff
@@ -1,10 +1,10 @@
 package bleep
 package rewrites
 
-case class keepSelectedProjects(selectedProjectGlobs: List[String]) extends BuildRewrite {
+case class keepSelectedProjects(selectedProjectGlobs: Array[model.ProjectGlob]) extends BuildRewrite {
```
sideeffffect, line 4: Why the changes to `Array`? If `List` is not good for some reason, why not at least `ArraySeq`?

sideeffffect (commented):
Looks awesome, thanks for this!
Just a note about Scala versions:
Scala used to have an epoch version. In Scala 2, `2` was the _epoch_ version. `2.13` was the _major_ version (and `2.13.10` was the minor version; there were no patch versions).
But this scheme changed. It is Semver now. With Scala 3, the `3` is the _major_ version, `3.2` is the _minor_ version and `3.2.2` is the _patch_ version.

So I don't think `@2` makes much sense. `@2` and `@3` is apples to oranges.
`@2.13` and `@3` is apples to apples.

Suggestions:
 * how about `@2.13` instead of `@213`?
 * It would be nice if we could (now or in the future) also do `@2.13.10`; such build matrices are required in some rare situations
 * What do you think about changing the schema to something more structured. We seem to have 3 axes (platform, versions and configuration), right? So something like `@js/2.13/test`? Or would `,` be a better symbol for separating the parts?

oyvindberg:
> Just a note about Scala versions: Scala used to have an epoch version. In Scala 2, `2` was the _epoch_ version. `2.13` was the _major_ version (and `2.13.10` was the minor version; there were no patch versions). But this scheme changed. It is Semver now. With Scala 3, the `3` is the _major_ version, `3.2` is the _minor_ version and `3.2.2` is the _patch_ version.
> So I don't think `@2` makes much sense. `@2` and `@3` is apples to oranges. `@2.13` and `@3` is apples to apples.

Yeah, I'm aware. but even without epoch `@3` is the string used to select all projects with binary version 3. So the only addition is that you're able to select all 2.x projects in one go. not sure if that pulls its weight really.


> 
> Suggestions:
> 
>     * how about `@2.13` instead of `@213`?

Maybe! The reason why I took out the `.` is that there is significant keytravel on the way from numbers to `.`, and the scala version selector is the one I use the most I think. It just feels more smooth without the `.` :)

> 
>     * It would be nice if we could (now or in the future) also do `@2.13.10`; such build matrices are required in some rare situations

This is actually somewhat supported. Full cross builds will receive crossIds like `jvm2.13.10`, so you'll be able to select them with `@jvm2.13.10`. I don't think there will ever be any full cross builds for other platforms so I'm not terribly concerned about that.

> 
>     * What do you think about changing the schema to something more structured. We seem to have 3 axes (platform, versions and configuration), right? So something like `@js/2.13/test`? Or would `,` be a better symbol for separating the parts?

Ok, I like the suggestion because I think it can remove the need for the `~` things to remove from selection.
Slightly more cumbersome to write maybe, but if we generate all relevant permutations of these things perhaps it'll be more clear and more flexible.

I'll give this a shot within to see how it feels.

I should note that the format chosen here comes from the generated crossIds (`js3`, `jvm213`), so to remain consistent those should probably change to something similar (`jvm/3`, `3,jvm`, `2.13/js`, `js,2.13`, or similar). In other words it's a "big" semantic change, but it's important to find something workable here

sideeffffect:
To make things even more complicated, we may want to add another axis, for the version of the platform -- Scala Native 0.4.x is not compatible with Scala Native 0.5.x. So that would make it something like `@native/0.4/2.13/test`... or just `@native0.4/2.13/test`...

We maybe should take inspiration from what is put in the name of the JAR and just do exactly that.
 * So for platform names it would be `native0.4` (instead of `native`) and `sjs1` (instead of `js`). `jvm` could be a special case which is "empty".
 * For Scala version numbers it would be `2.13`, not `213`.

https://repo1.maven.org/maven2/io/circe/circe-core_native0.4_2.13/
https://repo1.maven.org/maven2/io/circe/circe-core_sjs1_2.13/

sideeffffect:
Hi @oyvindberg , do you think it's worth reviving this?

PR #254, opened 2022-12-30
Schema: Allow RelPath to point to current directory

kluen:
This change allows using `./` as a relative path.

In a project I tried to convert, `bleep import` created a file with `templates.template-common.folder: ./`, which is correct, but was refused by the schema.


oyvindberg:
Thanks!

Issue #267, opened 2023-01-04
arm64 or aarch64 in the channel ? 


Baccata:
On my M1, calling `sys.props("os.arch")` returns `aarch64`

Coursier seems to just pass this information [through]((https://github.com/coursier/coursier/blob/c7f9658419e1fce68e5e0c222523ee8149175bd4/modules/install/src/main/scala/coursier/install/Platform.scala#L10) )  

Bleep however does refer to this architecture as [arm64](https://raw.githubusercontent.com/oyvindberg/bleep/master/coursier-channel.json)

This mistmatches leads coursier to attempt to dynamically rebuild the binary, which fails for an [unrelated reason](https://github.com/coursier/coursier/pull/2649#issuecomment-1370910025). 

I do wonder whether all Apple computers refer to the arch as `aarch64` or whether some of them refer to it as `arm64`. If the former is used, bleep should probably amend the channel, but I'm not 100% certain.

oyvindberg:
Yeah, this was a somewhat tough choice - there is usage of both `arm64` and `aarch64`. For instance coursier JVM downloads uses `arm64`. 

Bleep understands both and produces an ADT in [OsArch](https://github.com/oyvindberg/bleep/blob/master/bleep-model/src/scala/bleep/OsArch.scala#L31-L54). The idea I had is that the ADT does not contain a string version of the arch, but every usage of it should provide the proper arch strings it needs. For instance [FetchJvm](https://github.com/oyvindberg/bleep/blob/master/bleep-model/src/scala/bleep/FetchJvm.scala#L14-L17) and `FetchNode](https://github.com/oyvindberg/bleep/blob/master/bleep-model/src/scala/bleep/FetchNode.scala#L12-L18).

So if something is wrong in the coursier channel we should by all means fix that! :)
It's true that I've had limited/no success in making coursier download arm64 builds, so maybe you're onto something. So far I've attributes the failure to install an arm64 version to my coursier being built for amd64.

PR #271, opened 2023-01-05
Amend coursier channel so that cs finds aarch64-apple-darwin

Baccata:
Closes https://github.com/oyvindberg/bleep/issues/267

* Coursier is set to standardise on `aarch64` (see https://github.com/coursier/coursier/pull/2651/files) 
* I'm using an M1 and `sys.props("os.arch")` gives me `aarch64`, so I assume it's gonna be the same for all macs with an arm64 architecture, even in the current state without that coursier PR merged 

The fix was tried locally via the following command : 

``` 
cs install --channel file:///fully/qualified/path/to/coursier-channel.json bleep
```

which happens to work


oyvindberg:
fantastic, thanks ! :)

PR #282, opened 2023-01-31
Improve contributor experience

daddykotex:
Hi, thanks for this project. It's early days but I'm super excited by the potential and it's working great for us so far.

When attempting to contribute, I checkout the source but I could not run `bleep compile`, there were some errors regarding undefined types, like `bleep-model@jvm213 not found: type BleepVersion`

I looked at the CI and saw that you run `bleep generate-resources` before compiling, so I tried that.

I'm assuming this won't be required in the future, when things like https://github.com/oyvindberg/bleep/issues/19 are resolved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,10 @@ $ git clone --recurse-submodules https://github.com/oyvindberg/bleep.git
 See https://bleep.build/docs/installing/
 
 #### Use bleep to build bleep
-```
+```bash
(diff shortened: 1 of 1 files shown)


oyvindberg:
Absolutely, thanks!

This is indeed what #19 (implemented in #247) will fix. I hope to land it soon, just been busy with other things for some weeks

Issue #278, opened 2023-01-18
Bleep cache out of sync w/ manual JDK set up


lewisjkl:
Here is a minor issue that I ran into when getting up and running with Bleep.

1. Installed bleep with coursier (as docs describe)
2. Ran `bleep compile` and got a warning that helped me know my JDK version was too old (was using JDK 11)
3. Changed `$JAVA_HOME` to point to JDK 17
4. Re-ran `bleep compile`, but got the same error.
5. Realized that my JDK path was set explicitly to wrong version in `$HOME/.bloop/bloop.json` so updated that (not actually sure if this step is relevant to reproducing the issue)
6. Still no luck, so rebooted computer and still no change.
7. Eventually realized that bleep was caching the path in its cache, so `rm -rf ~/Library/Caches/build.bleep.bleep/` and then running `bleep compile` again ended up fixing the issue.

So I think the issue is basically, when using a manually managed JDK and you have the wrong version at first then bleep will (sometimes?) cache the path in a way that requires deleting the cache to get things working again.

Note: I have every intention of switching to the bleep-managed JDK, but wanted to post this in case it was relevant for others.

oyvindberg:
The for the report! I guess it caches the `system` jvm, that doesn't make sense to do :)

PR #283, opened 2023-01-31
Do not cache the system's jvm

daddykotex:
As seen in #278, caching the system's JVM can be tricky because updating JAVA_HOME will have no effect as the cache entry remains and the new JAVA_HOME is unused.

Closes https://github.com/oyvindberg/bleep/issues/278

Created as a draft because I'm not sure if you'd like test for this and I still need to manually test this here.


oyvindberg:
Thanks, seems straightforward to me :)

PR #293, opened 2023-02-12
Update UserPaths.scala

sideeffffect:
I'm not an expert on this, but looking at the example
https://github.com/dirs-dev/directories-jvm#example

With this proposed change you'd get
```
// Lin: /home/alice/.config/bleep
// Mac: /Users/Alice/Library/Application Support/build.bleep
// Win: C:\Users\Alice\AppData\Roaming\build\bleep\config
```
Which is probably the desired outcome?


sideeffffect:
Nah, I changed my mind about this... sorry :sweat_smile:

PR #296, opened 2023-02-13
Tweak error message for when script run fails

daddykotex:
This is just a small PR to tweak the error message that shows up when running a script that fails. Feel free to close it if you prefer the original, I just felt like `failed :(` was unhelpful


oyvindberg:
Absolutely! I'm very far from the level of polish where every message is carefully crafted just yet :)

PR #297, opened 2023-02-14
Notify bsp client upon build change

Baccata:
https://user-images.githubusercontent.com/4439335/218695611-7ed2f5f5-5238-4bea-9973-33c392621dad.mp4


oyvindberg:
More than awesome, thanks! Will look through the code hopefully tonight :)

bleep-core/src/scala/bleep/bsp/BleepBspServer.scala
```diff
@@ -141,7 +141,9 @@ class BleepBspServer(
   }
   override def buildTargetCompile(params: bsp4j.CompileParams): CompletableFuture[bsp4j.CompileResult] = {
     logger.debug(("buildTargetCompile", params.toString))
+    // buildChangeTracker {
```
daddykotex, line 144: letfover comment here?

daddykotex:
that's awesome, good job

oyvindberg:
Hey again, sorry this takes time.

I've started to look at it locally, but it's not working for me with neither vscode or intellij.

While debugging I keep thinking that the BSP integration is not very good (through no fault of yours!), especially error handling and logging. Maybe it's time to reconsider and invest in whatever you had in mind some time ago @Baccata 

So yeah, I'm looking further into this, but give me some time. Code-wise it looks all good, good work there.

bleep-core/src/scala/bleep/bsp/BuildChangeTracker.scala
```diff
@@ -0,0 +1,108 @@
...
+    maybeStarted match {
+      case Left(e)                  => State(pre, Left(e), SortedMap.empty)
+      case Right((newPre, started)) => State(newPre, Right(started), started.bloopProjects)
+    }
+  }
+
+  private def id(proj: Config.Project): BuildTargetIdentifier = {
+    val id = proj.directory.toUri.toASCIIString.stripSuffix("/") + "/?id=" + proj.name
+    // Applying the same format as bloop. There might be a better way to do this.
+    val amended = id.replace("file:///", "file:/")
```
oyvindberg, line 72: did this fix make any difference? I'm replacing usage of this method with [this existing method](https://github.com/oyvindberg/bleep/blob/master/bleep-core/src/scala/bleep/BleepCommandRemote.scala#L128) and I'm wondering if I should pick up this fix and apply it everywhere

Baccata: I _think_ it did yeah, but now I wonder if it's a platform specific kind of thing.

Baccata: And by platform I mean OS

oyvindberg:
It's working in metals by the way, and I'm wondering if intellij just doesn't have that feature implemented. 

Just so you know I'm working on top of this now, will push my changes to multiple PRs over the next days. Main topics are error handling over BSP (how to behave when reloading build and its broken, for instance), and I'm reimplementing sourcegen in bleep instead because it's just not working well enough. Will merge this more or less as-is as well in the coming days.

Baccata:
>  I'm reimplementing sourcegen in bleep instead because it's just not working well enough 

I've been meaning to report that we've trialed it and seen some very weird behaviour leading to non-deterministic compile errors on unrelated things. Haven't found time to create a minimal repro, unfortunately.

oyvindberg:
Yeah, too many things for me to bother patching bloop this time around, also I'm fairly sure there is a very old compiler issue around generated sources which makes the compiler see them twice and fail. I'm hoping that the different timing we get from generating the sources a bit earlier may be enough to fix it. So far looks good.

oyvindberg:
> I mean the ideal approach in my mind would be swapping bsp4j for my fs2-based [jsonrpclib](https://github.com/neandertech/jsonrpclib) library so that concurrency can be controlled and unit-tested easily, and generating the BSP interface from smithy once my [BSP PR](https://github.com/build-server-protocol/build-server-protocol/pulls) is merged/released.

> There's still a lot to do there before I can even start trialing this approach in my bleep fork, and chances are you may be able to evolve things in a way that you'd find preferable before I have all the pieces lined up.

That has the potential to fantastic, really eager to see how it goes :)
I'll suffer with bsp4j in the meantime, but let's definitely explore this :)

PR #298, opened 2023-02-14
Do not exit on error when no error occurred

daddykotex:
Today, I realized that running `bleep --help` gives a non-zero exit code. This is a fix for that.

### DIFF
--- a/bleep-cli/src/scala/bleep/Main.scala
+++ b/bleep-cli/src/scala/bleep/Main.scala
@@ -516,6 +516,9 @@ object Main {
 
   def run(logger: Logger, opts: Opts[BleepNoBuildCommand], restArgs: List[String]): ExitCode =
     Command("bleep", s"Bleeping fast build! (version ${model.BleepVersion.current.value})")(opts).parse(restArgs, sys.env) match {
+      case Left(help @ Help(Nil, _, _, _)) =>
+        System.err.println(help)
+        ExitCode.Success
       case Left(help) =>
         System.err.println(help)
         ExitCode.Failure
--- a/bleep-cli/src/scala/bleep/Main.scala
+++ b/bleep-cli/src/scala/bleep/Main.scala
@@ -502,31 +502,38 @@ object Main {
   }
 
   def run(logger: Logger, opts: Opts[BleepBuildCommand], started: Started, restArgs: List[String]): ExitCode =
-    Command("bleep", s"Bleeping fast build! (version ${model.BleepVersion.current.value})")(opts).parse(restArgs, sys.env) match {
-      case Left(help) =>
-        System.err.println(help)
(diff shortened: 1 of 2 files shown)


sideeffffect (approved):
I'd even do it as
```scala
case Left(List()) =>
  System.err.println(help)
  ExitCode.Success
```
But this surely works too :+1:

daddykotex:
I can amend my PR, no worries.

daddykotex:
I'm having trouble building locally, not sure if my changes are the cause of this.

oyvindberg:
Looks good, thanks!

What do you think is the problem @daddykotex ? if you can paste output perhaps I can help diagnose it - compilation should work! :)

daddykotex:
I did not spend much time investigating, but it looks like this:

```

**Comment by oyvindberg** [comment]
Have you checked out git submodules?

**Comment by daddykotex** [comment]
Ahhh it's related to the submodule. Got it to work with `git submodule update --remote`

**Review by sideeffffect — COMMENTED** [review]
I see that the different branches are implemented differently. Is that intentional?
(I don't have very strong preference either way.)

**Comment by daddykotex** [comment]
> I see that the different branches are implemented differently. Is that intentional? (I don't have very strong preference either way.)

Woww, it's my bad. Sorry I was convinced I was updating the same code path not two different codepaths! I've updated with a common `handleHelp` method that I use on a `fold` on the two `Either[Help, A]` we get from command parsing.

wdyt?

**Comment by daddykotex** [comment]
That's a great improvement, I used it and reordered the args at the call site following your changes

File: bleep-cli/src/scala/bleep/Main.scala [hunk 1]
```diff
@@ -502,28 +502,38 @@ object Main {
...
+            case Success(Right(_)) => ExitCode.Success
+          }
+      )
+
+  private def handleHelp(help: Help): ExitCode =
+    help.errors match {
+      case Nil =>
+        System.err.println(help)
+        ExitCode.Success
+      case _ => // should we print the errors?
```
oyvindberg, line 535: I think `System.err.println(help)` takes care of it? You can try to println them to verify what is in there

oyvindberg (approved):
Nice improvement, thanks! :)

sideeffffect:
@daddykotex I'm glad you like my proposal :bow: Have you already commited and pushed your changes?

daddykotex:
Yes. The only left over is whether or not we should print the errors
included in the help object if any.

Its possible that they're already printed, I'll validate and update my code
later tonight and it will be ready to merge after that.

bleep-cli/src/scala/bleep/Main.scala
daddykotex: You're right, the toString implementation print errors if any

daddykotex:
It's good to go.

PR #299, opened 2023-02-17
Main.scala: consolidate run methods

sideeffffect:


bleep-cli/src/scala/bleep/Main.scala
```diff
@@ -501,39 +501,25 @@ object Main {
...
-      .fold(
-        handleHelp,
-        cmd =>
-          Try(cmd.run(started)) match {
-            case Failure(th)       => fatal("command failed", logger, th)
-            case Success(Left(th)) => fatal("command failed", logger, th)
-            case Success(Right(_)) => ExitCode.Success
-          }
-      )
+  def run(logger: Logger, restArgs: List[String], opts: Opts[BleepBuildCommand], started: Started): ExitCode =
```
oyvindberg, line 504: inline these two `run` methods too maybe?

sideeffffect: done

PR #302, opened 2023-02-24
Rewrite sourcegen so we run it from bleep. Stabilize and polish

oyvindberg:


oyvindberg:
Things seem to work much, much better now. 

The way I see it now there is only one problem left. When working with intellij and you update the build file and/or project selection, bleep will automatically reload it while intellij won't. A desync like this may obviously cause some grief. Bleep may for instance refuse to compile if you deselected a project which is still active in intellij.

I'll open a separate issue for that, not sure what to do about it for now.

PR #313, opened 2023-05-26
Bump versions, scala to 3.3.0

oyvindberg:
the big diffs are from changing the constants which affect new and imported builds


oyvindberg:
I think this will leave scripts written in scala 3.x, where x < 3 in the strange situation where 2.13 jars work better than 3 jars

PR #327, opened 2023-06-09
Add exclusions to GenLayout for maven

dispalt:
TODO:

- [ ] Deal with ivy as well

### DIFF
--- a/bleep-core/src/scala/bleep/packaging/GenLayout.scala
+++ b/bleep-core/src/scala/bleep/packaging/GenLayout.scala
@@ -79,10 +79,10 @@ object GenLayout {
       </publications>
       <dependencies>{
       deps.map { dep =>
-        <dependency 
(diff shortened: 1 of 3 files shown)


oyvindberg:
Great, thanks!

Give me a hint if you want to skip ivy, like I said on gitter I'm hopeful we can remove it entirely. 

Otherwise I'll merge it right away :)

dispalt:
> Great, thanks!
> 
> Give me a hint if you want to skip ivy, like I said on gitter I'm hopeful we can remove it entirely. 
> 
> Otherwise I'll merge it right away :)

Yeah let's nuke it. I tested the maven version locally but maybe we can write a more substantial test with coursier.

oyvindberg:
yeah we could definitely look into testing this more in depth. open another PR if you want to dig in that direction :)

PR #330, opened 2023-06-14
set bloop working directory to project root

frawa:
Setting the build servers working directory to be the project's root folder.

This is useful for macros resolving files relative to the project root.

Should this choice be configurable?

### DIFF
--- a/bleep-cli/src/scala/bleep/bsp/BspImpl.scala
+++ b/bleep-cli/src/scala/bleep/bsp/BspImpl.scala
@@ -43,7 +43,8 @@ object BspImpl {
           resolvedJvm = pre.resolvedJvm.forceGet,
           userPaths = pre.userPaths,
           resolver = resolver,
-          bleepRifleLogger = bleepRifleLogger
+          bleepRifleLogger = bleepRifleLogger,
+          workingDir = Some(pre.buildPaths.cwd)
         )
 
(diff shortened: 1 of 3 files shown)


oyvindberg:
Thanks for your interest in solving this! :) Unfortunately this way won't work. The bloop compile server is a process which runs across bleep invocations and across different builds. 

To try to add some detail:
Say you first compile project `a1` in build file `$HOME/a/bleep.yaml`, and then compile project `b1` in build file `$HOME/b/bleep.yaml`. Say bloop wasn't running when you compiled `a1`, so bleep would start a bloop server with working directory set to say `$HOME/a`. if you then compile `b` it'll still have that working directory set.

frawa:
> Thanks for your interest in solving this! :) Unfortunately this way won't work. The bloop compile server is a process which runs across bleep invocations and across different builds.
> 
> To try to add some detail: Say you first compile project `a1` in build file `$HOME/a/bleep.yaml`, and then compile project `b1` in build file `$HOME/b/bleep.yaml`. Say bloop wasn't running when you compiled `a1`, so bleep would start a bloop server with working directory set to say `$HOME/a`. if you then compile `b` it'll still have that working directory set.

Indeed, I was able to confirm: Different projects share the same bloop server. The first project to start the server wins with its cwd. Making the other fail.

Issue #341, opened 2023-07-18
bleep import fails unexpectedly


kpodsiad:
Hey @oyvindberg! I wanted to try bleep, but I have encountered problem when importing sbt build. I can take a look at problem, but is there some sort of contributing guide which describes basic worfklow when working with bleep itself? I'll start playing with it anyway in case you didn't manage to read this issue. Have a good day!
<details><summary>Stack trace</summary>

<p>

```
 command failed unexpectedly java.nio.charset.MalformedInputException: Input length = 1
        at java.base@17.0.6/java.lang.String.throwMalformed(String.java:1238)
        at java.base@17.0.6/java.lang.String.decodeUTF8_UTF16(String.java:1189)
        at java.base@17.0.6/java.lang.String.newStringUTF8NoRepl(String.java:726)
        at java.base@17.0.6/java.lang.String.newStringNoRepl1(String.java:754)
        at java.base@17.0.6/java.lang.String.newStringNoRepl(String.java:736)
        at java.base@17.0.6/java.lang.System$2.newStringNoRepl(System.java:2390)
        at java.base@17.0.6/java.nio.file.Files.readString(Files.java:3369)
        at java.base@17.0.6/java.nio.file.Files.readString(Files.java:3325)
        at bleep.sbtimport.ImportInputData$.findFiles$1$$anonfun$2$$anonfun$2(ImportInputData.scala:160)
        at scala.collection.Iterator$$anon$9.next(Iterator.scala:584)
        at scala.collection.Iterator$$anon$10.next(Iterator.scala:618)
        at scala.collection.immutable.List.prependedAll(List.scala:156)
        at scala.collection.immutable.List$.from(List.scala:684)
        at scala.collection.immutable.List$.from(List.scala:681)
        at scala.collection.IterableOps$WithFilter.flatMap(Iterable.scala:901)
        at bleep.sbtimport.ImportInputData$.findFiles$1(ImportInputData.scala:160)
        at bleep.sbtimport.ImportInputData$.$anonfun$17(ImportInputData.scala:163)
        at scala.collection.StrictOptimizedIterableOps.flatMap(StrictOptimizedIterableOps.scala:118)
        at scala.collection.StrictOptimizedIterableOps.flatMap$(StrictOptimizedIterableOps.scala:105)
        at scala.collection.immutable.Vector.flatMap(Vector.scala:113)
        at bleep.sbtimport.ImportInputData$.collectFromFileSystem(ImportInputData.scala:164)
[...]

PR #342, opened 2023-07-19
fix: don't fail when file reading binary files during import

kpodsiad:
closes https://github.com/oyvindberg/bleep/issues/341


kpodsiad:
Interesting, `bleep compile bleep-tests` fails with 
```
📕 BSP: bleep-tests missing argument for parameter logger of method collectFromFileSystem in object ImportInputData: (destinationPaths: bleep.BuildPaths, logger: bleep.logging.Logger):
  bleep.sbtimport.ImportInputData [value => 171]
```
and there is `171` instead of filename

oyvindberg:
thank you!

PR #357, opened 2023-09-12
Move help output form stderr to stdout

SlowBrainDude:
I've never seen a CLI tool outputting help messages on stderr until now. Looks wrong to me.

Also improved the failure message with some more end-user actionable suggestion. Users couldn't see the code comment… :smile:

### DIFF
--- a/bleep-cli/src/scala/bleep/Main.scala
+++ b/bleep-cli/src/scala/bleep/Main.scala
@@ -405,7 +405,7 @@ object Main {
         }
     }
 
-  def main(_args: Array[String]): Unit = {
+  def _main(_args: Array[String]): ExitCode = {
     val userPaths = UserPaths.fromAppDirs
 
     val exitCode: ExitCode = _args.toList match {
@@ -485,9 +485,11 @@ object Main {
           }
         }
     }
-
-    System.exit(exitCode.value)
+    exitCode
   }
+
+  def main(args: Array[String]): Unit =
+    System.exit(_main(args).value)
 
   def withCompletions(_args: Array[String], userPaths: UserPaths, commonOpts: CommonOpts, args: List[String])(f: Completer.Res => ExitCode): ExitCode = {
     val (_, restArgs) = CommonOpts.parse(args)
@@ -537,10 +539,13 @@ object Main {
   private def run[Cmd](opts: Opts[Cmd], restArgs: List[String], logger: Logger)(runCommand: Cmd => Either[BleepException, Unit]): ExitCode =
     Command("bleep", s"Bleeping fast build! (version ${model.BleepVersion.current.value})")(opts).parse(restArgs, sys.env) match {
       case Left(help) =>
-        System.out.println(help)
         help.errors match {
-          case List() => ExitCode.Success
-          case _      => ExitCode.Failure
+          case List() =>
+            System.out.println(help)
+            ExitCode.Success
+          case _ =>
+            System.err.println(help)
+            ExitCode.Failure
         }
       case Right(cmd) =>
         Try(runCommand(cmd)) match {
--- a/bleep-cli/src/scala/bleep/Main.scala
+++ b/bleep-cli/src/scala/bleep/Main.scala
@@ -537,14 +537,14 @@ object Main {
   private def run[Cmd](opts: Opts[Cmd], restArgs: List[String], logger: Logger)(runCommand: Cmd => Either[BleepException, Unit]): ExitCode =
     Command("bleep", s"Bleeping fast build! (version ${model.BleepVersion.current.value})")(opts).parse(restArgs, sys.env) match {
       case Left(help) =>
-        System.err.println(help)
+        System.out.println(help)
         help.errors match {
           case List() => ExitCode.Success
           case _      => ExitCode.Failure
         }
       case Right(cmd) =>
         Try(runCommand(cmd)) match {
-          case Failure(th)        => fatal("command failed unexpectedly", logger, th) // This really shouldn't happen
+          case Failure(th)        => fatal("command failed unexpectedly! This really shouldn't happen. Please report.", logger, th)
           case Success(Left(th))  => fatal("command failed", logger, th)
           case Success(Right(())) => ExitCode.Success
         }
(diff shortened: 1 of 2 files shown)


bleep-cli/src/scala/bleep/Main.scala
```diff
@@ -537,14 +537,14 @@ object Main {
   private def run[Cmd](opts: Opts[Cmd], restArgs: List[String], logger: Logger)(runCommand: Cmd => Either[BleepException, Unit]): ExitCode =
     Command("bleep", s"Bleeping fast build! (version ${model.BleepVersion.current.value})")(opts).parse(restArgs, sys.env) match {
       case Left(help) =>
-        System.err.println(help)
```
sideeffffect, line 540: I thought this is actually quite common. std-out/-in is used the data the program is supposed to work on. stderr is supposed to contain diagnostic out, like errors or command line arguments help.

SlowBrainDude: Could you show an example? I'm wondering.

sideeffffect: image

https://pubs.opengroup.org/onlinepubs/009695399/

sideeffffect: GNU guidelines say, that if the user explicitly asked for `--help`, it should go to stdout. But if the help text was printed because of badly configured program, it should go to stderr.

https://www.gnu.org/prep/standards/html_node/_002d_002dhelp.html

SlowBrainDude: The `stderr` is indeed for diagnostics data (it doesn't need to be "errors"). I knew this.

But I've tried actually with `--help` in some tools. All output than to stdout…

Now the question is: Is this code change too stupid to make this work correctly? Will this code path be triggered also for failed run attempts? Let me check this, as I'm not sure!

At least we seem to agree on what to implement. :smile:  I'll take the GNU recommendation as a guideline.

Thanks for your assistance!

oyvindberg: Fully in agreement that it should go to stdout on errors. 

apologies for late feedback on your work @SlowBrainDude - it's been a crazy few last days with scala days :)

SlowBrainDude: I've refined the logic according to the GNU standards (normal `--help` goes on stdout, but not when it's part of failed run attempts, than it goes to stderr like the rest).

SlowBrainDude:
I've added also tests for the refined logic.

This test cases could be expended in the future for other CLI IO tests that needs to go through `main` like this one.

oyvindberg (approved):
Looks good, thanks!

SlowBrainDude:
Pushed again after running `bleep fmt`… Sorry for this oversight.

SlowBrainDude:
I've tried to pick up the suggestion from the comment on the other PR and pass the IO Streams as "resources" to the new "fake main" instead of doing the switching in the test implementation.

But I guess I don't fully understand the idea.

There are two ways to do it as I see it:

You can just move `System.setOut(outputStream)` and `System.setErr(errorStream)` to `_main`. and get the *Stream params from params given to `_main`. This way you could pass `System.out` and `System.err` in the real `main`, or the buffers when calling the fake `_main` from the tests.

But imho there is not much to win this way. You will still redirect the JVM std IO streams. Instead encapsulated in the tests (as this is not related to "normal" function of the app) now it's in `Main` and influences the design. Needing to change your code just for tests is imho a smell. (And I had to do this already because of `System.exit`…)

The other approach would be to treat system IO streams as proper resources and pass them down where they're used. Treating IO streams as reassure is not really feasible on the JVM, imho. The code reads and prints to the std. streams directly usually. You have a call to `println` just a few lines into `_main` actually… So the only way to redirect that is to redirect the whole std. IO of the JVM again—or encapsulating every call to any function that does std. IO (like `println`) into some resource manager… The later is just not realistic. At least if you don't go 110% into some effect system, or so. (Which I hope won't happen as this would make things needlessly complex and likely slow!)

If you like I can move of course the calls to `System.setOut(outputStream)` and `System.setErr(errorStream)`  to `_main` and path stuff through params. But like said I'm not even sure this is what you had in mind. (And like said, I don't see much value in it actually.)

oyvindberg:
I'm happy to merge this as-is, suggest we do so :)

As for the last comment:


> The other approach would be to treat system IO streams as proper resources and pass them down where they're used. 
> But imho there is not much to win this way. You will still redirect the JVM std IO streams

Oh yeah, I absolutely meant as parameters. I use the compiler as an added test suite, and that means parameters in this case.

It's not a big reactoring. Feel free to grep the codebase - there are very few mentions of `println`, `System.out` and `System.err`. Basically in 
- the BSP code (which integrates over these streams). I suggest we don't touch that at all, as I'm hoping to replace the whole thing with another stack once it materializes.
- in `Main` (because these streams have to be those streams there). 
- in a few build diff commands where I wanted more "plain" output. This is obviously a work in progress.
- and most importantly in the setup of the loggers. 

*All other logging* is done through loggers - bleep can even send log statements as json across processes for consistent rendering when it calls out to scripts.


> Needing to change your code just for tests is imho a smell
100% opposite of how I see it. writing tests is how you get well designed code. writing code without tests first is how you get code written.

SlowBrainDude:
> I'm happy to merge this as-is, suggest we do so :)

Cool. :+1: 

> I use the compiler as an added test suite, and that means parameters in this case.

I'm not sure what you mean.

But I agree that passing the streams as "resources" would be the cleaner design. Would make things more "pure".

> It's not a big reactoring. Feel free to grep the codebase

I need to take a look. Not sure what it would mean to pass the streams explicitly. But OK, I don't know this codebase…

> writing code without tests first

Let's put it like that: I have a more nuanced opinion on that. But this would be a really long write-up and completely off-topic.

What is good is that this PR brings a way to test the output of this thingy end-to-end without needing to spawn external processes. I was missing this facility in the other PR already… Having such kind of tests is really important!

oyvindberg:
> I'm not sure what you mean.
oh sorry, the context was that the compiler can track params, but it cannot track the side-effect of overriding stdout globally

> Let's put it like that: I have a more nuanced opinion on that. But this would be a really long write-up and completely off-topic.
😆 

> What is good is that this PR brings a way to test the output of this thingy end-to-end without needing to spawn external processes. I was missing this facility in the other PR already… Having such kind of tests is really important!

👍

PR #360, opened 2023-09-20
Update schema.json to define "index" as a "string" variable.

megri:
I think string is the correct type here? It seems like boolean definitely is the wrong one at least 😁


oyvindberg:
absolutely! thanks! :)

PR #376, opened 2024-05-16
fix: always include bleep as cmd for zsh tab completions

### LINKED ISSUE #375: incorrect zsh compleation for coursier installations
Observed on Ubnuntu. 

**Reproduce:**
1. Install bleep using coursier as deceived here: https://bleep.build/docs/installing/
2. Call `bleep install-tab-completions-zsh`
3. Refresh the shell

**Expected:**
auto-completion on the `bleep` command

**Actual:**
auto-completion on the `bleep` does not work. It does however work when using the binary from coursier `.bleep.aux`. 

**Discovery:**
Coursier does wrap the bleep binary (`.bleep.aux`) in a shell script (`bleep`). The command to generate the completion resolves to the binary over the wrapper script. As a temporary workaround I modified the `~/.config/bleep/zsh/_.bleep.aux` script to include the bleep command.

```diff
-#compdef _.bleep.aux .bleep.aux
+#compdef _.bleep.aux .bleep.aux bleep
```

**How to fix it?**
1. Ensure we always add bleep to the compleation as done in the workaround. Can cause issues if we have multiple versions bleep installed and want to keep them separated. 
2. Solve issues with coursier installations: Include some hint about the runtime in [BleepExecutable.scala#L26] . There's already som special handling of java runtime at [BleepExecutable.scala#L74].
3. Generic fix for wrapper scripts: Need more investigation. 

[BleepExecutable.scala#L26]: https://github.com/oyvindberg/bleep/blob/master/bleep-core/src/scala/bleep/BleepExecutable.scala#L26
[BleepExecutable.scala#L74]: https://github.com/oyvindberg/bleep/blob/master/bleep-core/src/scala/bleep/BleepExecutable.scala#L74

ingarabr:
Resolves #375

### DIFF
--- a/bleep-cli/src/scala/bleep/commands/InstallZshTabCompletions.scala
+++ b/bleep-cli/src/scala/bleep/commands/InstallZshTabCompletions.scala
@@ -10,7 +10,8 @@ import java.nio.file.Path
 case class InstallZshTabCompletions(userPaths: UserPaths, logger: Logger) extends BleepCommand {
   override def run(): Either[BleepException, Unit] = {
     val programName = BleepExecutable.findCurrentBleep(logger) match {
-      case Some(binary: BleepExecutable.Binary) => binary.command.getFileName.toString
+      case Some(CoursierInstallation(scriptPath, _)) => scriptPath.getFileName.toString
+      case Some(binary: BleepExecutable.Binary)      => binary.command.getFileName.toString
       case _ =>
         logger.warn("Doesn't know name of a Bleep executable. Falling back to 'bleep'")
         "bleep"
--- a/bleep-cli/src/scala/bleep/commands/CoursierInstallation.scala
+++ b/bleep-cli/src/scala/bleep/commands/CoursierInstallation.scala
@@ -1,0 +1,21 @@
+package bleep.commands
+
+import bleep.BleepExecutable
+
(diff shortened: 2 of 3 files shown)


bleep-cli/src/scala/bleep/commands/InstallZshTabCompletions.scala
```diff
@@ -15,9 +15,11 @@ case class InstallZshTabCompletions(userPaths: UserPaths, logger: Logger) extend
         logger.warn("Doesn't know name of a Bleep executable. Falling back to 'bleep'")
         "bleep"
     }
+    val completionCommands =
+      if (programName != "bleep") List(programName, "bleep") else List(programName)
```
sideeffffect, line 19: Maybe pattern matching will make it super clear what's going on?

oyvindberg:
Thanks for this!

Can this be done more explicitly? In the case around line 13, we have found a binary. Check if the name of the binary ends with `.aux`, check if a file with the same path minus the suffix exists, and return that if it does. Comment that this is a workaround for coursier doing weird things.

I don't want to add `bleep` unconditionally, because this mechanism can be used to test two differently named bleep binaries, including completions.

Also can we do the same for bash?

oyvindberg:
perfect, thanks!

PR #382, opened 2024-06-03
Adds option for user provided path to sbt binary for bleep import command.

KristianAN:
The current bleep import command can and will not work on systems such as NixOs and Guix, so in my effort of making bleep play nice with my system I propose to add this option that lets the user provide the path to the launcher binary. 

This way NixOs/Guix users are then able to provide their own sbt instances with the correct utils in scope to actually run the bleep import.

I have given this a spin and it works great!

The only caveat is that for e.g. nix flakes you need to somewhat know what you are doing and add an overlay to sbt to provide the correct tools.

```nix
 sbt = pkgs.sbt.overrideAttrs (oldAttrs: {
    nativeBuildInputs = oldAttrs.nativeBuildInputs or [] ++ [ pkgs.makeWrapper ];
    postFixup = ''
      wrapProgram $out/bin/sbt --prefix PATH : "${pkgs.coreutils}/bin:${pkgs.gnugrep}/bin:${pkgs.gnused}/bin"
    '';
  });
 ```
 
 If this PR (or a derivative) is merged I'll update my bleep-flake repo with the above snippet.
 https://github.com/KristianAN/bleep-flake/pull/3/files

### DIFF
--- a/.github/prepare-release.sh
+++ b/.github/prepare-release.sh
@@ -12,5 +12,8 @@ chmod +x artifacts/bleep-x86_64-pc-linux/bleep
 chmod +x artifacts/bleep-x86_64-apple-darwin/bleep
 (cd artifacts/bleep-x86_64-apple-darwin && tar cvfz "$CWD/work/release/bleep-x86_64-apple-darwin.tar.gz" bleep)
 
+chmod +x artifacts/bleep-arm64-apple-darwin/bleep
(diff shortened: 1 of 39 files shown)


KristianAN:
@oyvindberg If this is merged I'll publish bleep to nixpkgs as it seems like all the main features work (only completions that aren't supported yet). Once published I can update the docs here and remove the reference to my bleep.nix flake and instead maybe add a few lines of how to use with nix and how to get the sbt import working.

oyvindberg:
Perfect, thanks. I'm offline all this weekend, but aiming for a release early next week 👍

PR #385, opened 2024-06-07
Added flag to send shell completions to stdout.

KristianAN:
This will allow users that have bespoke systems to pipe and modify the completions as they see fit, increasing the number of systems shell completions can be installed too.


oyvindberg:
Thanks.

 It's rather hard for me to see why this is useful, but I'll just assume it's a nix thing?

PR #391, opened 2024-06-24
Added support for setting ModuleSplitStyleJS

KristianAN:
Waiting for: https://github.com/scala-cli/bloop-core/pull/167 + release

Creating this PR before the change in bloop-core is merged to get some feedback on the way I have handled this for the bleep.yaml schema. This was the best I could come up with. Feedback appreciated.

A possible alternative is to define two objects, one for the split stlyle and one for packages and then just join those two internally, allowing a flatter structure in the bleep.yaml. The default option used by esmodules will probably be SmallestModules anyway.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,6 +94,11 @@ jobs:
 #        if: runner.os == 'Windows'
 
       - name: Build native image 2 (windows)
+        run: bleep compile bleep-cli@jvm3
+        shell: cmd
+        if: runner.os == 'Windows'
+
+      - name: Build native image 3 (windows)
         uses: graalvm/setup-graalvm@v1
         if: runner.os == 'Windows'
         with:
--- a/.gitmodules
+++ b/.gitmodules
@@ -48,8 +48,3 @@
     url = git@github.com:bleep-build/librarymanagement.git
     branch = develop
     update = rebase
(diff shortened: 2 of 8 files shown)


oyvindberg (approved):
Thanks! Model changes look uncontroversial to me. 

Have you tested the json schema in an IDE? I'm not sure I understand all of it, but if intellij understands it I'm happy

KristianAN:
Tested the schema using a scalajs bleep project and Intellij was autocompleting and linting like a champ. choosing esmodules now gives linter warning about missing required field.

```yaml
  template-js:
    platform:
      jsEmitSourceMaps: false
      jsJsdom: false
      jsKind: esmodule
      jsMode: debug
      jsNodeVersion: 18.4.0
      jsVersion: 1.16.0
      jsSplitStyle:
        splitStyle: FewestModules
      name: js
```

if splitStyle: SmallModulesFor is set it gives a warning about missing required property packages

KristianAN:
I also think it's possible to set regex rules for the packages property in the schema. There is probably many such things that could be added here and there to the schema to provide better linting. There is obviously the trade-off that it makes the schema a lot more complex. It might be one of those places where the complexity is worth it though.

KristianAN:
The required change is merged into scalacenter/bloop. So we can hold off on this until bleep uses that instead of the fork

PR #394, opened 2024-06-28
--test-only

KristianAN:
So this is a very crude implementation that should work once https://github.com/scalacenter/bloop/pull/2360 is merged and bleep uses mainline bloop. 

However the cli interface is not great as it will look like bleep test project -o TestSuiteOne -o TestSuiteTwo ... and so on. I don't think it's possible to combine Option and arguments in decline? So I'm guessing here we have to use arguments and just do some parsing to make it look nice?

Update:

It is now possible to specify which test suites to run using the -o or
--only syntax.

The tests can be specified using the class name of the test suite or by
using the fully qualified name.

Example:
bleep test app-tests -o TestSuiteOne --only TestSuiteTwo
fully qualified:
bleep test app-tests -o com.foo.TestSuiteThree

### DIFF
--- a/bleep-core/src/scala/bleep/commands/Test.scala
+++ b/bleep-core/src/scala/bleep/commands/Test.scala
@@ -8,6 +8,7 @@ import scala.jdk.CollectionConverters.*
 
 import bloop.rifle.BuildServer
 import cats.data.NonEmptyList
+import bleep.BleepException.TestSuitesNotFound
 
 case class Test(watch: Boolean, projects: Array[model.CrossProjectName], testSuites: Option[NonEmptyList[String]])
     extends BleepCommandRemote(watch)
@@ -36,26 +37,35 @@ case class Test(watch: Boolean, projects: Array[model.CrossProjectName], testSui
 
       val testParams = new bsp4j.TestParams(targets)
 
-      testSuites.foreach { classes =>
-        val testClassesData = new bsp4j.ScalaTestSuites(
-          intersectTestSuites(testClasses, classes.toList)
-            .map(cls => new bsp4j.ScalaTestSuiteSelection(cls, List.empty[String].asJava))
-            .asJava,
-          List.empty[String].asJava,
-          List.empty[String].asJava
-        )
-        testParams.setData(testClassesData)
-        testParams.setDataKind(bsp4j.TestParamsDataKind.SCALA_TEST_SUITES_SELECTION)
-      }
+      val hasSuites: Boolean = testSuites
+        .map { classes =>
+          val testClassesData = new bsp4j.ScalaTestSuites(
+            intersectTestSuites(testClasses, classes.toList)
+              .map(cls => new bsp4j.ScalaTestSuiteSelection(cls, List.empty[String].asJava))
+              .asJava,
+            List.empty[String].asJava,
+            List.empty[String].asJava
+          )
+          testParams.setData(testClassesData)
+          testParams.setDataKind(bsp4j.TestParamsDataKind.SCALA_TEST_SUITES_SELECTION)
 
-      val result = bloop.buildTargetTest(testParams).get()
+          !testClassesData.getSuites().isEmpty
+        }
+        .getOrElse(true)
 
-      result.getStatusCode match {
-        case bsp4j.StatusCode.OK =>
-          started.logger.info("Tests succeeded")
-          Right(())
-        case errorCode =>
-          Left(new BspCommandFailed("tests", projects, BspCommandFailed.StatusCode(errorCode)))
+      if (hasSuites) {
+        val result = bloop.buildTargetTest(testParams).get()
+
+        result.getStatusCode match {
+          case bsp4j.StatusCode.OK =>
+            started.logger.info("Tests succeeded")
+            Right(())
+          case errorCode =>
+            println(errorCode)
+            Left(new BspCommandFailed("tests", projects, BspCommandFailed.StatusCode(errorCode)))
+        }
+      } else {
+        Left(new TestSuitesNotFound(testSuites.map(_.toList).getOrElse(List.empty[String])))
       }
     }
 }
--- a/bleep-core/src/scala/bleep/commands/Test.scala
+++ b/bleep-core/src/scala/bleep/commands/Test.scala
@@ -31,6 +31,69 @@ case class Test(
       }
     }
 
+  private def runTests(started: Started, bloop: BuildServer, testParams: bsp4j.TestParams) = {
+    val result = bloop.buildTargetTest(testParams).get()
+
+    result.getStatusCode match {
+      case bsp4j.StatusCode.OK =>
+        started.logger.info("Tests succeeded")
+        Right(())
+      case errorCode =>
+        Left(new BspCommandFailed("tests", projects, BspCommandFailed.StatusCode(errorCode)))
+    }
+  }
+
+  private def runWithFilters(started: Started, bloop: BuildServer, targets: java.util.List[bsp4j.BuildTargetIdentifier]) = {
+    val tests = bloop.buildTargetScalaTestClasses(new bsp4j.ScalaTestClassesParams(targets)).get()
+    val testClassesUnfiltered: List[String] = tests.getItems().asScala.flatMap(_.getClasses.asScala).toList
+
+    val testClasses = testSuitesExclude.map { exclude =>
+      val filtered = testClassesUnfiltered
+        .filter(fromBuildCls => exclude.toList.find(fromCliCls => isSameSuite(fromBuildCls, fromCliCls)).isDefined)
+      if (filtered.nonEmpty) testClassesUnfiltered.diff(filtered) else List.empty
+    }
+
+    val testParams = new bsp4j.TestParams(targets)
+
+    val hasSuites: Boolean =
+      testClasses match {
+        case None =>
+          testSuitesOnly
+            .map { classes =>
+              val testClassesData = new bsp4j.ScalaTestSuites(
+                intersectTestSuites(testClassesUnfiltered, classes.toList)
+                  .map(cls => new bsp4j.ScalaTestSuiteSelection(cls, List.empty[String].asJava))
+                  .asJava,
+                List.empty[String].asJava,
+                List.empty[String].asJava
+              )
+              testParams.setData(testClassesData)
+              testParams.setDataKind(bsp4j.TestParamsDataKind.SCALA_TEST_SUITES_SELECTION)
+
+              !testClassesData.getSuites().isEmpty
+            }
+            .getOrElse(true)
+        case Some(value) =>
+          if (value.isEmpty) false
+          else {
+            val testClassesData = new bsp4j.ScalaTestSuites(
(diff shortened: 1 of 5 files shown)


bleep-core/src/scala/bleep/commands/Test.scala
```diff
@@ -18,7 +22,22 @@ case class Test(watch: Boolean, projects: Array[model.CrossProjectName]) extends
...
-      val result = bloop.buildTargetTest(new bsp4j.TestParams(targets)).get()
+      val tests = bloop.buildTargetScalaTestClasses(new bsp4j.ScalaTestClassesParams(targets)).get()
+      val testClasses: List[String] = tests.getItems().asScala.flatMap(_.getClasses.asScala).toList
+
+      val testParams = new bsp4j.TestParams(targets)
+
+      testSuites.foreach { classes =>
+        val testClassesData = new bsp4j.ScalaTestSuites(
+          testClasses.filter(classes.toList.contains).map(cls => new bsp4j.ScalaTestSuiteSelection(cls, List.empty[String].asJava)).asJava,
+          List.empty[String].asJava,
```
KristianAN, line 33: Consider allowing passing jvm options

bleep-core/src/scala/bleep/commands/Test.scala
```diff
@@ -18,7 +22,22 @@ case class Test(watch: Boolean, projects: Array[model.CrossProjectName]) extends
...
       val targets = BleepCommandRemote.buildTargets(started.buildPaths, projects)
-      val result = bloop.buildTargetTest(new bsp4j.TestParams(targets)).get()
+      val tests = bloop.buildTargetScalaTestClasses(new bsp4j.ScalaTestClassesParams(targets)).get()
+      val testClasses: List[String] = tests.getItems().asScala.flatMap(_.getClasses.asScala).toList
+
+      val testParams = new bsp4j.TestParams(targets)
+
+      testSuites.foreach { classes =>
+        val testClassesData = new bsp4j.ScalaTestSuites(
+          testClasses.filter(classes.toList.contains).map(cls => new bsp4j.ScalaTestSuiteSelection(cls, List.empty[String].asJava)).asJava,
```
KristianAN, line 32: So this is pretty crude as you can have foo.TestSuite and bar.TestSuite and it if you set TestSuite to run it will run both, but you could always just say foo.TestSuite and it would only run that test. Thisis because ScalaTestClassesParams returns the fully qualified class name.

bleep-core/src/scala/bleep/commands/Test.scala
```diff
@@ -18,7 +22,22 @@ case class Test(watch: Boolean, projects: Array[model.CrossProjectName]) extends
   override def runWithServer(started: Started, bloop: BuildServer): Either[BleepException, Unit] =
     DoSourceGen(started, bloop, watchableProjects(started)).flatMap { _ =>
       val targets = BleepCommandRemote.buildTargets(started.buildPaths, projects)
-      val result = bloop.buildTargetTest(new bsp4j.TestParams(targets)).get()
+      val tests = bloop.buildTargetScalaTestClasses(new bsp4j.ScalaTestClassesParams(targets)).get()
```
KristianAN, line 25: This is deprecated, but I am not quite sure how the JVM replacement works. I will look into it.

oyvindberg:
Great start! I'll follow bloop development and switchover once there is a new release of mainstream 👍

KristianAN:
@oyvindberg See the updated top comment. This works, I have not written any tests for it, but I tried it on a dummy project. The question remaining is the CLI interface. I personally don't mind the -o syntax for each suite, in practice I never use this functionality for more than one suite at a time anyway, but opinions may differ. Different syntax is not well supported by decline and requires more manual parsing on our end.

KristianAN:
Also if no matching test is found it runs all tests. This might not be desired and can easily be changed to no tests are run if the build has no suite matching the one passed in using --only

KristianAN:
Another fun and maybe useful thing we could add here is an --ignore or --not which could ignore test suites. Could be useful if you want to e.g. ignore some slow test but not move it to a separate test module

bleep-core/src/scala/bleep/commands/Test.scala
oyvindberg: this has been that way for a while. we can do this in a follow-up I think

oyvindberg:
> Also if no matching test is found it runs all tests. This might not be desired and can easily be changed to no tests are run if the build has no suite matching the one passed in using --only

Can we change this to what you suggested? I would be irritated if i tried to use `--only` and bleep would just start to run everything without any further feedback

bleep-core/src/scala/bleep/commands/Test.scala
oyvindberg: I'd like a follow-up task where we standardize handling of environment variables, java properties and so on between `test`,`run`, `script` (maybe others?) - and then document it properly

oyvindberg: see #406

bleep-core/src/scala/bleep/commands/Test.scala
```diff
@@ -4,10 +4,23 @@ package commands
...
+case class Test(watch: Boolean, projects: Array[model.CrossProjectName], testSuites: Option[NonEmptyList[String]])
+    extends BleepCommandRemote(watch)
+    with BleepCommandRemote.OnlyChanged {
+
+  private def intersectTestSuites(fromBuild: List[String], fromCli: List[String]): List[String] =
+    fromBuild.filter { fullyQualifiedCls =>
+      fromCli.exists { fromCliCls =>
+        val fromBuildFqn = fullyQualifiedCls.split('.')
+        val fromCliFqn = fromCliCls.split('.')
+        fromBuildFqn.sameElements(fromCliFqn) || fromBuildFqn.last == fromCliFqn.last
```
oyvindberg, line 21: I don't really like this predicate here. 

to start: can we make both lists, `fromBuild` and `fromCli` lowercase before we start filtering?
then let's have one rule instead of two. one of the following maybe:

- check if `fromCliFqn` is a substring of  `fromBuildFqn` 
- let `fromCliFqn` be a regex and see if it matches
- let `fromCliFqn` be a glob and see if it matches
or a rule like that.

then document it, at least with a description in the CLI interface

bleep-core/src/scala/bleep/commands/Test.scala
```diff
@@ -18,7 +22,22 @@ case class Test(watch: Boolean, projects: Array[model.CrossProjectName]) extends
...
+      val tests = bloop.buildTargetScalaTestClasses(new bsp4j.ScalaTestClassesParams(targets)).get()
+      val testClasses: List[String] = tests.getItems().asScala.flatMap(_.getClasses.asScala).toList
+
+      val testParams = new bsp4j.TestParams(targets)
+
+      testSuites.foreach { classes =>
+        val testClassesData = new bsp4j.ScalaTestSuites(
+          testClasses.filter(classes.toList.contains).map(cls => new bsp4j.ScalaTestSuiteSelection(cls, List.empty[String].asJava)).asJava,
+          List.empty[String].asJava,
+          List.empty[String].asJava
```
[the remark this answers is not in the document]
oyvindberg: I guess it's not possible to list all test in test suites before running? I think this is less important

oyvindberg:
> Another fun and maybe useful thing we could add here is an --ignore or --not which could ignore test suites. Could be useful if you want to e.g. ignore some slow test but not move it to a separate test module

this is a great idea. maybe it could take the form of `--include` / `-i` (instead of `-o`) or `--exclude` / `-x` for test suites maybe?

bleep-core/src/scala/bleep/commands/Test.scala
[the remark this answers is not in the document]
KristianAN: I'm not 100% sure what bloop actually provides through bsp here. I can take a look when I have time.

KristianAN:
> > Another fun and maybe useful thing we could add here is an --ignore or --not which could ignore test suites. Could be useful if you want to e.g. ignore some slow test but not move it to a separate test module
> 
> this is a great idea. maybe it could take the form of `--include` / `-i` (instead of `-o`) or `--exclude` / `-x` for test suites maybe?

I think the --only -o syntax is good because it is close to what scala-cli, bloop and sbt uses, the --exclude -x seems very intuitive. I actually think this could be really useful for CI stuff as you could exclude tests that don't run on certain platforms and so on.

bleep-core/src/scala/bleep/commands/Test.scala
KristianAN: I initially went with the substring approach, but after testing it some I really did not like it as I felt it became too easy for tests to clash, e.g RepositoryTest is a substring of IntegrationRepositoryTest. So my idea here was to make it as predictable as possible. It either runs whatever suites has that name e.g AppTest will run foo.AppTest and bar.AppTest and if you want to pick one you specify with fully qualified name. 

I will add documentation to the CLI and bleep.build page once the predicate is decided :)

KristianAN:
> > Also if no matching test is found it runs all tests. This might not be desired and can easily be changed to no tests are run if the build has no suite matching the one passed in using --only
> 
> Can we change this to what you suggested? I would be irritated if i tried to use `--only` and bleep would just start to run everything without any further feedback

Fixed
If no tests-suites in the project match the specified test suites
provided by the user no tests will run at all. If any of the tests
match, but not all the matching tests will run.

KristianAN:
> > Another fun and maybe useful thing we could add here is an --ignore or --not which could ignore test suites. Could be useful if you want to e.g. ignore some slow test but not move it to a separate test module
> 
> this is a great idea. maybe it could take the form of `--include` / `-i` (instead of `-o`) or `--exclude` / `-x` for test suites maybe?

I have added support for this in the latest commit. It's a bit iffy as the -x and -o options are mutually exclusive, but this is not enforced by the CLI.

oyvindberg:
ok, let's test this now with the current predicate. we can always iterate more later

bleep-core/src/scala/bleep/commands/Test.scala
```diff
@@ -18,14 +40,62 @@ case class Test(watch: Boolean, projects: Array[model.CrossProjectName]) extends
...
       val targets = BleepCommandRemote.buildTargets(started.buildPaths, projects)
-      val result = bloop.buildTargetTest(new bsp4j.TestParams(targets)).get()
-
-      result.getStatusCode match {
-        case bsp4j.StatusCode.OK =>
-          started.logger.info("Tests succeeded")
-          Right(())
-        case errorCode =>
-          Left(new BspCommandFailed("tests", projects, BspCommandFailed.StatusCode(errorCode)))
+      val tests = bloop.buildTargetScalaTestClasses(new bsp4j.ScalaTestClassesParams(targets)).get()
```
oyvindberg, line 43: ok, one other thing. can you adjust this method so that `buildTargetScalaTestClasses` is only ran if if one or both of `testSuitesOnly` or `testSuitesExclude` are set? It would be a shame to make the default case slower by calling bloop twice

PR #400, opened 2024-08-01
Switched bloop back to scalacenter/bloop

KristianAN:
Preparing to switch back to scalacenter/bloop so we can start merging in the pending PRs.

### DIFF
--- a/bleep-core/src/scala/bleep/bsp/SetupBloopRifle.scala
+++ b/bleep-core/src/scala/bleep/bsp/SetupBloopRifle.scala
@@ -41,17 +41,17 @@ object SetupBloopRifle {
     model.LibraryVersionScheme.VersionScheme.Always,
     model.Dep.Scala("org.scala-lang.modules", "scala-parallel-collections", "always")
   )
-  val versionCombo = model.VersionCombo.Jvm(model.VersionScala.Scala213)
+  val versionCombo = model.VersionCombo.Jvm(model.VersionScala.Scala212)
 
-  def bloopClassPath(resolver: CoursierResolver)(bloopVersion: String): Either[BleepException, (Seq[File], Boolean)] = {
-    val dep = model.Dep.Scala("io.github.alexarchambault.bleep", "bloop-frontend", bloopVersion)
+  def bloopClassPath(resolver: CoursierResolver)(bloopVersion: String): Either[BleepException, Seq[File]] = {
+    val dep = model.Dep.Scala("ch.epfl.scala", "bloop-frontend", bloopVersion)
     resolver
       .updatedParams(_.copy(downloadSources = false))
       .resolve(Set(dep), versionCombo, libraryVersionSchemes = SortedSet(parallelCollectionAlways)) match {
       case Left(coursierError) =>
         Left(new BleepException.ResolveError(coursierError, "installing bloop"))
       case Right(value) =>
-        Right((value.jarFiles, true))
+        Right(value.jarFiles)
     }
   }
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -94,7 +94,7 @@ jobs:
         if: runner.os == 'Windows'
 
       - name: Build native image 2 (windows)
-        run: bleep config compile bleep-cli@jvm3
+        run: bleep compile bleep-cli@jvm3
         shell: cmd
         if: runner.os == 'Windows'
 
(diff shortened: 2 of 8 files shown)


KristianAN:
@oyvindberg Maybe you have some insight into why the windows build fails?

oyvindberg:
No, unfortunately. Im considering disabling the windows build soon, until a motivated windows user comes along. I have no idea whether the windows port actually works now before this PR.

bleep-core/src/scala/bleep/bsp/SetupBloopRifle.scala
```diff
@@ -41,17 +41,17 @@ object SetupBloopRifle {
     model.LibraryVersionScheme.VersionScheme.Always,
     model.Dep.Scala("org.scala-lang.modules", "scala-parallel-collections", "always")
   )
-  val versionCombo = model.VersionCombo.Jvm(model.VersionScala.Scala213)
+  val versionCombo = model.VersionCombo.Jvm(model.VersionScala.Scala212)
```
oyvindberg, line 44: They only publish it for 2.12?!

oyvindberg (approved):
Nice! Will merge when we get it green

oyvindberg:
I'll try to do some debug magic tomorrow or Sunday night if you can't find a way for windows.

bleep-core/src/scala/bleep/bsp/SetupBloopRifle.scala
KristianAN: It would seem so. I'm not really sure why. Probably it just has not been a priority while the fork was maintained?

KristianAN:
> I'll try to do some debug magic tomorrow or Sunday night if you can't find a way for windows.

I would not even know where to start as I don't have access to a windows machine. It's worth noting that the windows CI seems to be very flaky for scalacenter/bloop as well. 

Examples:
https://github.com/scalacenter/bloop/actions/runs/9900282981
https://github.com/scalacenter/bloop/actions/runs/9893468259
https://github.com/scalacenter/bloop/actions/runs/9892903253

bleep-core/src/scala/bleep/bsp/SetupBloopRifle.scala
sideeffffect: > They only publish it for 2.12?!

Is that really so @tgodzik?

Also, why are you using `bloop-frontend` when scala-cli is using `bloop-rifle`?

Relevant PRs
 * https://github.com/scalacenter/bloop/pull/2355
 * https://github.com/VirtusLab/scala-cli/pull/3020

tgodzik: But looking here, I think that's fine. It's just the internal version of the server. I think we might be able to update to 2.13 now, but didn't have the chance to look.

oyvindberg: No idea where the frontend dependency comes from, it should all be rifle, just like scala-cli

tgodzik: That's the dependency for starting Bloop server itself. Bloop rifle doesn't download anything by itself.

KristianAN: This is because bleep bundles bloop instead of expecting the user to have it on PATH right?

KristianAN: Doesn't scala-cli and metals have to do something similar? (metals might not need to as it can depend on the build tool to manage it). If the user does not have bloop on PATH you will need bloop-frontend to start the server, no?

tgodzik: We never look for Bloop on PATH. If it was already started then we can just use the same artifacts it downloaded, otherwise we just download what is needed and provide classpath to bloop rifle. But yes, the current approach here is the same as Scala CLI and Metals do it.

tgodzik (approved):
From my perspective everything here is expected, not sure about the windows issues. The ones in Bloop are normally related to DAP timing out.

oyvindberg:
tried some cleanups of earlier hacks to get the windows build going. so far we're blocked on this thing:
```
📙 Failed to delete C:\Users\runneradmin\AppData\Local\bleep\cache\socket\proc-196 at shutdown: C:\Users\runneradmin\AppData\Local\bleep\cache\socket\proc-196\output: The process cannot access the file because it is being used by another process
```

I think I struggled with this the last update as well. the problem is fundamentally that we're building bleep with and old bloop, then using the new bleep which requires a new bloop, and it's struggling to shut down the old bloop server. this has been a recurring issue honestly. maybe it's time to really figure out how to fix it

oyvindberg:
ok, I just disabled the auto-shutdown feature on windows (which was used to work around earlier problems). now it's running with old bleep/bloop throughout. good enough for now.

PR #402, opened 2024-08-03
Support `compileServerOptions` for when you need to compile with a javac with options set

oyvindberg:
You'll get a bloop server with these flags set, isolated from those with same version but other flags.

You specify the options in yaml:
```
jvm:
  name: ...
  compileServerOptions:
    - foo
    - bar
```

You can also set `compileServerOptionIncludeDefault = false` if your needs conflict with the default flags from bloop


oyvindberg:
This works, and lets me work with experimental JVM options and so on, which is awesome.

I'll let it linger a bit more, because I'm wondering if these flags should be propagated to the normal `javacOptions` in all the projects. Gonna have a to investigate a bit

PR #403, opened 2024-08-05
Adds support for modulekindsplitstyle

KristianAN:
This is the same as:
https://github.com/oyvindberg/bleep/pull/391
Which I might or might not have pushed from my master, which I also might or might not have accidentaly overwritten.

I need some guidance with what is going on in the failing test here.


oyvindberg:
this is the failing test. maybe there is an `@unchecked` somewhere which allows non-exhaustive pattern matches?

```
📗 BSP: - bleep.CreateNewSnapshotTests:
📗 BSP:   * create-new-build - scala.MatchError: Js (of class bleep.model.PlatformId$Js$)
```

KristianAN:
That would live in bloop then? If so I will dig into it later tonight sometime

oyvindberg:
no, in bleep. maybe in test code. you should be able to get a stack trace for that `MatchError` somehow

KristianAN:
> no, in bleep. maybe in test code. you should be able to get a stack trace for that `MatchError` somehow

It was nothing but a simple bug in my code. Fixed now.

oyvindberg:
I'll go ahead and merge this now. we'll fix in followups if anything is wrong

PR #407, opened 2024-08-08
HOTFIX - Fix to logic in schema valiation

KristianAN:
Last commit introduced a slight bug where it would complain about missing jsSplitStyle even when it should not be set. e.g it would complain for
platform:
  name: jvm

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -291,27 +291,16 @@
       "dependencies": {
         "jsSplitStyle": ["jsKind"]
       },
-      "anyOf": [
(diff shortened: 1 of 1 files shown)


KristianAN:
Please squash on merge :octocat:

KristianAN:
To avoid these kinds of things in the future we could create a CI action with https://github.com/InoUno/yaml-ls-check?

oyvindberg:
fancy! :)

PR #410, opened 2024-08-08
Watcher fix

oyvindberg:


KristianAN:
My CPU will be so happy, and my SSD will live on

PR #413, opened 2024-08-10
housekeeping:

oyvindberg:
- remove most unused warnings
- introduce `bleep.Discard` to do so
- bump bleep
- optimize imports
- fmt
- enforce `.model` prefix

### DIFF
--- a/bleep-cli/src/scala/bleep/commands/InstallZshTabCompletions.scala
+++ b/bleep-cli/src/scala/bleep/commands/InstallZshTabCompletions.scala
@@ -32,7 +32,7 @@ case class InstallZshTabCompletions(userPaths: UserPaths, logger: Logger, stdout
       println(completionScript)
     } else {
       logger.info(s"Writing $completionScriptDest")
-      FileSync.softWriteBytes(completionScriptDest, completionScript.getBytes(StandardCharsets.UTF_8)): Unit
+      FileSync.softWriteBytes(completionScriptDest, completionScript.getBytes(StandardCharsets.UTF_8)).discard()
       val zshRc = Option(System.getenv("ZDOTDIR")).map(Path.of(_)).getOrElse(FileUtils.Home) / ".zshrc"
 
       PatchRcFile(None, logger, zshRc)(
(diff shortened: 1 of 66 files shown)


oyvindberg:
there is a mountain of unused warnings for xml literals for scala 3 still, doesn't seem fixable unless rewritten into something else.

KristianAN:
yes, the warnings for xml literals is something we also have a lot of at $work. I have not found a workaround yet, and there does not seem to be much activity to fix it.

oyvindberg:
yeah, also couldn't find anything except some old issues. I think the plan for the last decade has been to remove the xml literals and replace with string interpolators which nobody has written yet

bleep-cli/src/scala/bleep/commands/InstallZshTabCompletions.scala
```diff
@@ -32,7 +32,7 @@ case class InstallZshTabCompletions(userPaths: UserPaths, logger: Logger, stdout
       println(completionScript)
     } else {
       logger.info(s"Writing $completionScriptDest")
-      FileSync.softWriteBytes(completionScriptDest, completionScript.getBytes(StandardCharsets.UTF_8)): Unit
+      FileSync.softWriteBytes(completionScriptDest, completionScript.getBytes(StandardCharsets.UTF_8)).discard()
```
sideeffffect, line 35: Why not just
```scala
val _ = FileSync.softWriteBytes(completionScriptDest, completionScript.getBytes(StandardCharsets.UTF_8))
```
There's no need to invent custom extension methods to do a simple thing like this...

oyvindberg: Believe it or not, cross compiling between strict 2.13 and 3 provides no way to resolve all unused cases. And you need to use 3-4-5 different syntaxes to resolve most unused cases. It's laughable, really. Inventing this is way easier to deal with

sideeffffect: Oh boy :scream_cat: fair enough then...

PR #428, opened 2024-09-23
Updated BuildUpdateDeps to use the Version algorithm from Scala Steward

KristianAN:
Adds support for upgrading using the algorithm from Scala Steward in the simplest way. 

basically it's just this that gets us the latest version
```scala
  def selectLatest(versions: List[Version], allowPreReleases: Boolean = false): Option[Version] = {
    def recur(current: Version): Version =
      current.selectNext(versions, allowPreReleases) match {
        case None        => current
        case Some(value) => recur(value)
      }

    selectNext(versions, allowPreReleases).map(recur)
  }
 ``` 
Added a flag for using the same upgrade strategy as scala steward and a flag for allowing prerelease
  
Setting this to draft because I have not gotten prerelease to work yet. When I tested it with the milestone release of http4s it just went into an infinite loop, when on M41 it says the next version is M40 and when on M40 the next is 41.

### CODE REVIEW

File: bleep-cli/src/scala/bleep/internal/Version.scala [hunk 1]
```diff
@@ -0,0 +1,238 @@
+/*
```
oyvindberg, line 1: I guess this file is copy/pasted from scala steward? are there any changes? on other words: is there any need to review anything in here?

oyvindberg (approved):
Nice! If it's hard to get preleases going I would be ok with shipping this without it. it's incredibly rare to use milestone releases, and if you do the consequence would just be that this particular command doesn't work optimally.

bleep-cli/src/scala/bleep/internal/Version.scala
KristianAN: The snippet in the top comment is added to it. And I removed this
implicit val versionCodec: Codec[Version] = deriveUnwrappedCodec
As it uses circe-generic-extras and I don't see any reason we need the codec.

KristianAN:
I might have an idea for an ok solution to the prerelease for now

oyvindberg:
this ready to merge you think? I haven't had the time to test it yet, but will before I cut a release with this and #429

KristianAN:
Maybe wait until I have written the tests for #429. Might be able to add some tests for this while I am at it.

PR #429, opened 2024-09-23
Adds possibility of upgrading a single dependency

KristianAN:
followup to #428 

can now pass --single <dep> to update-deps

it's possible to pass only an org or an org + module
e.g. passing org.http4s will upgrade all deps of the org.http4s
organization, while passing org.http4s::http4s-dsl will only upgrade
that dependency.

allows for an arbitrary amount amount of colon symbols between org and module.


KristianAN:
This could also be it's own subcommand update-dep with the same flags as update-deps, thoughts?

bleep-cli/src/scala/bleep/commands/BuildUpdateDeps.scala
```diff
@@ -32,6 +39,25 @@ case class BuildUpdateDeps(scalaStewardMode: Boolean, allowPrerelease: Boolean)
       Await.result(fetchAllVersions(fileCache, repos, allDeps), Duration.Inf)
     }
 
+    singleDepParsed match {
```
oyvindberg, line 42: can you extract all of the logic here into an `object` and write some tests for it?

oyvindberg (approved):
Thanks, this is a fantastic feature.

I'm wondering if this should be a new command? `bleep build update-dep` vs `bleep build update-deps` for instance. what do you think?

oyvindberg:
hah, I see now you already proposed it

KristianAN:
I also closed #428 as this PR cleans up the code quite a bit

KristianAN:
Specifying a single dependency is at the moment extremely lenient in that it will accept anything before :: as the org and anything after as the module name. This is mostly because I have no idea what ivy/maven actually allows here.

bleep-cli/src/scala/bleep/Main.scala
```diff
@@ -122,6 +122,12 @@ object Main {
 
     val watch = Opts.flag("watch", "start in watch mode", "w").orFalse
 
+    val updateAsScalaSteward = Opts.flag("steward", "Use same upgrade strategy as Scala Steward").orFalse
+
+    val updateWithPrerelease = Opts.flag("prerelease", "Allow upgrading to prerelease version if there is any.").orFalse
+
+    val updateSingleOrgOrModule = Opts.argument[String]("The depenency to update, alternatively only the organization name can be passed").orNone
```
KristianAN, line 129: This should maybe not be orNone?

oyvindberg: I think it's fine, that it makes it optional

bleep-cli/src/scala/bleep/Main.scala
```diff
@@ -122,6 +122,12 @@ object Main {
 
     val watch = Opts.flag("watch", "start in watch mode", "w").orFalse
 
+    val updateAsScalaSteward = Opts.flag("steward", "Use same upgrade strategy as Scala Steward").orFalse
+
+    val updateWithPrerelease = Opts.flag("prerelease", "Allow upgrading to prerelease version if there is any.").orFalse
+
+    val updateSingleOrgOrModule = Opts.argument[String]("The depenency to update, alternatively only the organization name can be passed")
```
oyvindberg, line 129:
```suggestion
    val updateSingleOrgOrModule = Opts.argument[String]("The dependency to update, alternatively only the organization name can be passed")
```

bleep-cli/src/scala/bleep/Main.scala
```diff
@@ -122,6 +122,12 @@ object Main {
 
     val watch = Opts.flag("watch", "start in watch mode", "w").orFalse
 
+    val updateAsScalaSteward = Opts.flag("steward", "Use same upgrade strategy as Scala Steward").orFalse
```
oyvindberg, line 125: are you able to briefly explain what the strategy is? bump minor to newest, and if minor is newest then pick newest major?

bleep-cli/src/scala/bleep/commands/BuildUpdateDeps.scala
```diff
@@ -31,12 +36,16 @@ case object BuildUpdateDeps extends BleepBuildCommand {
...
-        val latest = version.latest
-        if (latest == bleepDep.version) None
-        else Some(tuple -> bleepDep.withVersion(latest))
-      }
+    DependencyUpgrader.depsToUpgrade(singleDep, foundByDep, scalaStewardMode, allowPrerelease).flatMap { toUpdate =>
+      runUgrades(started, build, toUpdate)
+    }
+  }
+
+  private def runUgrades(
```
oyvindberg, line 44: *updates

oyvindberg (commented):
This looks very good now. I left a few comments for some typos.

One thing I thought about - java dependencies. does it work to `update-dep` them as well? Maybe it works, but in case you didn't think about it: scala dependencies are specified by `org::module:version` (double `:`), while java dependencies are specified by `org:module:version` (single `:`).

KristianAN:
Yeah it works for Java dependencies as well.

nafg:
My 2 cents, one command `update-deps` is fine, IIUC it would take an optional pattern, if that matches only a single maven coordinate that's just a special case of "update deps." I think fewer subcommands where reasonable is better.

Besides, the same maven coordinates can be depended on by multiple subprojects, so even exact coordinates can be updating multiple dependencies, since arguably each occurrence is a separate dependency ("dependency" means a case of X depending on Y, so it's a different dependency if it's a different Y or a different X...)

KristianAN:
> Besides, the same maven coordinates can be depended on by multiple subprojects, so even exact coordinates can be updating multiple dependencies, since arguably each occurrence is a separate dependency ("dependency" means a case of X depending on Y, so it's a different dependency if it's a different Y or a different X...)

We can handle this by making the command interactive or take a list of module names to upgrade (no list meaning all modules). This can probably be autocompleted as well.

nafg:
> > Besides, the same maven coordinates can be depended on by multiple subprojects, so even exact coordinates can be updating multiple dependencies, since arguably each occurrence is a separate dependency ("dependency" means a case of X depending on Y, so it's a different dependency if it's a different Y or a different X...)
> 
> We can handle this by making the command interactive or take a list of module names to upgrade (no list meaning all modules). This can probably be autocompleted as well.

To clarify, it was an argument to justify the plural form. Of course it's unimportant bike shedding...

nafg:
BTW probably OT but...

I don't think Scala Steward supports Bleep, nor Renovate or Dependabot AFAIK (not sure what's the most appropriate -- SS is needed where the build is code, if the build is YAML it's probably very easy to add support to the others, but I suppose all would be nice...)

I'm just wondering if the work in this PR is at all relevant to making that happen.

KristianAN:
> BTW probably OT but...
> 
> I don't think Scala Steward supports Bleep, nor Renovate or Dependabot AFAIK (not sure what's the most appropriate -- SS is needed where the build is code, if the build is YAML it's probably very easy to add support to the others, but I suppose all would be nice...)
> 
> I'm just wondering if the work in this PR is at all relevant to making that happen.

I'm not sure how relevant this is. I have a branch on SS where I have started adding support for updating bleep builds, but I have no eta on when I will have it done.

nafg:
Cool, amazing

oyvindberg:
finally had the time to test this now. looks good! I fixed a merge conflict and force pushed, will merge on green 👍

PR #445, opened 2024-12-15
Adds a link command

KristianAN:
This is useful for projects where linking is a part of another toolchain such as a JS project where we might link a ScalaJS project that uses JSExport and then build the resulting files with e.g. vite.

this is blocked by: https://github.com/scalacenter/bloop/pull/2535

This PR also adds support for the newest configuration options in bloop-config that are supported by bloop for linking SN


oyvindberg (approved):
Thanks! This has been missing for so long, happy you could work on it!

KristianAN:
With the changes to in my PR to bloop we can specify if we want release or debug mode dynamically through BSP. So this begs the question, for linking do we want two commands; linkDebug/linkRelease. Or do we want link with flags (--debug/--release) to override what is set in the model? users coming from sbt will be used to defining the mode through the command. Personally I like having one command with flags that override the model the best, this is explicit in the build definition, yet flexible in use. 

Note that Bloop does not (yet at least) support changing the linker mode through for the Run command.

bleep.yaml
```diff
@@ -20,8 +20,7 @@ projects:
       mainClass: bleep.Main
   bleep-core:
     dependencies:
-    - for3Use213: true
-      module: ch.epfl.scala::bloop-config:2.1.0
+    - ch.epfl.scala::bloop-config:2.2.0
```
KristianAN, line 23: bloop-config is now published for scala 3

KristianAN:
Added support for overriding the linker mode by passing either(--release -r or --debug -d flags when linking.

oyvindberg (approved):
Fantastic. The only thing I wanted to comment is the `release`/`debug` names. I suppose you used those in bloop already? Because I would have expected them to mirror the fast/full optimization names from elsewhere. Maybe it's good enough to drop that in comments. Thoughts?

Otherwise lgtm

KristianAN:
I just used those as it's what bloop defines as the liker mode. Cargo uses a --release flag as well when linking with optimizations. I do like the release flag, but I agree that we could add some more information in the CLI help about it being equivalent to fastLinkJS (not sure what scala-native calls this). I don't like the debug name as much, but I don't think that will be used that much in any case? What I am guessing will be the common way to use this is setting the linker mode to debug in bleep.yaml then overriding in the cli with --release.

oyvindberg:
> Cargo uses a --release flag as well when linking with optimizations
I'm fine with following cargo's lead here.

> What I am guessing will be the common way to use this is setting the linker mode to debug in bleep.yaml then overriding in the cli with --release.

this got me thinking, is there ever a need to specify linking mode in the bleep/bloop projects? it would be a bit less confusing for users if we just removed that from the model, bloop would always default to `debug`, and the only way you can link in `release` mode would be to add the flag to `bleep link`? wdyt?

nafg:
Indeed, there aren't debug projects and release projects, but in general you want to link them in debug mode locally while developing and in the CI deployment pipeline you want to link in release mode. 

BTW I think scala-native also uses debug/release terminology? As well as C# in Visual Studio, I think.

KristianAN:
I have now changed it so that bleep defaults to debug for linking and the linker mode option is removed from the bleep model. The user can override when linking by passing the --release flag. 

One caveat is that this means that we currently can only call run on a native/js target with debug as --release is not supported as an argument in bloop. This isn't really an issue as 99% of the time when you run a native binary/js project from bloop you want it to be in debug anyway. However I think it will be worth it to add the same functionality for run as for link in this case. So when I have time I will add this to bloop and to run here.

oyvindberg:
Thank you! This will be great! 

I'm completely ok with run always being in default mode for now. It's not very hard to first link and then run if you have the need

oyvindberg:
Just un-draft it when you're ready to merge

sideeffffect:
Maybe we can ask @WojciechMazur about what is the best terminology. We probably don't want to introduce a new, bleep-proprietary lingo, different from what Scala Native and its docs use.

nafg:
It looks like Scala Native has 3 modes https://scala-native.org/en/stable/user/sbt.html#link-time-optimization-lto

nafg:
It looks like Scala Native uses "debug"/"release" but it actually has a few release variants: https://scala-native.org/en/stable/user/sbt.html#compilation-modes

sideeffffect:
I don't know if it's relevant here, but the bloop-config's `NativeConfig` seems quite out of sync with the most recent Scala Native Config

 * https://github.com/scalacenter/bloop-config/blob/c41b5a229cc9bae0a3ae6e252e0cefd6846eb0d1/config/src/bloop/config/Config.scala#L210-L232
 * https://github.com/scala-native/scala-native/blob/9e7db82ab78c0ccd310c425e8b17ece5705a6870/tools/src/main/scala/scala/scalanative/build/NativeConfig.scala#L314-L315

For example the `lto` is missing and `mode` is there, but doesn't have the correct values. Should somebody fix this?

sideeffffect:
It seems like Bloop is simplifying this waaay too much, almost to the point of making things up...
 * https://github.com/scalacenter/bloop/blob/c1ff013f47867902c13ba69b01c41c7f0ff76180/bridges/scala-native-0.5/src/main/scala/bloop/scalanative/NativeBridge.scala#L40-L48

sideeffffect:
I got a bit off topic, but my main point is that Scala Native uses these terms
 * `buildTarget` (`application`, `libraryDynamic`, `libraryStatic`)
 * `mode` (`debug`, `release-fast`, `release-size`, `release-full`)
 * `lto` (`none`, `thin`, `full`)

So I'd suggest we stick to using these and nothing else.

KristianAN:
Yes, there are still quite a few discrepancies when it comes to linking SN with bloop (probably some with JS as well). My plan was/is to fix this after we have a good way of linking.

KristianAN:
> I got a bit off topic, but my main point is that Scala Native uses these terms
> 
>     * `buildTarget` (`application`, `libraryDynamic`, `libraryStatic`)
> 
>     * `mode` (`debug`, `release-fast`, `release-size`, `release-full`)
> 
>     * `lto` (`none`, `thin`, `full`)
> 
> 
> So I'd suggest we stick to using these and nothing else.

As link is also for ScalaJS I don't think we should have too much going on in the CLI. I think --release is just fine. ScalaJS only has one release mode, where the optimizations are determined by other options. I also don't want to pollute the BSP compile arguments too much with things we could have in bloop-config. If we had a remote interface only for linking then adding more parameters would probably be fine. 

I propose that we use the --release flag as described in this PR. Then when I have had time to bring bloop-config and the bloop native bridge up to speed we can add more options to the bleep model. That way we can set sane defaults and give the user the option to override in the config. I.e. --release will default to release-fast, but let the user be able to set releaseMode: size, or something similar. This way we can keep our CLI interface stable and evolve the bleep model whenever there are changes to scala native. 

Thoughts?

oyvindberg:
👍 for keeping terminology as close as possible, especially for SN.

I think one point which is difficult is to determine which of these settings are properties of the project (that is, should go in the model), and which are properties of the linking you want to do right now (should be given to the `link` command). I firmly believe that release/debug belongs to the latter - for the rest of the SN options I have no opinion, really.

how does scala-cli solve this for SN? can you specify everything inline as directives and override at link-time?

KristianAN:
> how does scala-cli solve this for SN? can you specify everything inline as directives and override at link-time?

Yes this is how it works for scala-cli. However, this is not something bloop supports yet. The simplest working solution now is to keep debug/release tied to the CLI and the rest in the model. Once  bloop has better support for changing these settings through BSP we can look at expanding the CLI to support more. 

I will make sure that we use the same terminology in the bloop model that SN uses.

bleep-model/src/scala/bleep/model/VersionScalaNative.scala
```diff
@@ -16,7 +16,18 @@ case class VersionScalaNative(scalaNativeVersion: String) {
 
   val testInterface = Dep.ScalaDependency(VersionScalaNative.org, ModuleName("test-interface"), scalaNativeVersion, fullCrossVersion = false)
   val scalaLib = Dep.ScalaDependency(VersionScalaNative.org, ModuleName("scalalib"), scalaNativeVersion, fullCrossVersion = false)
-  val scala3Lib = Dep.ScalaDependency(VersionScalaNative.org, ModuleName("scala3lib"), scalaNativeVersion, fullCrossVersion = false)
+  def scala3Lib(scalaVersion: String) =
```
KristianAN, line 19: This solved #448

bleep-model/src/scala/bleep/model/Platform.scala
```diff
@@ -24,9 +23,15 @@ case class Platform(
     //        classpath: Option[List[Path]],
     //        resources: Option[List[Path]]
     nativeVersion: Option[VersionScalaNative],
-    nativeMode: Option[LinkerMode],
     nativeGc: Option[String],
-    nativeBuildTarget: Option[NativeBuildTarget]
+    nativeBuildTarget: Option[NativeBuildTarget],
```
KristianAN, line 27: bloop-config supports more options, but they aren't (yet) used by bloop.

bleep-model/src/scala/bleep/model/NativeBuildTarget.scala
```diff
@@ -5,24 +5,24 @@ import io.circe.*
 sealed abstract class NativeBuildTarget(val id: String)
 object NativeBuildTarget {
   case object Application extends NativeBuildTarget("application")
-  case object LibraryDynamic extends NativeBuildTarget("dynamic")
-  case object LibraryStatic extends NativeBuildTarget("static")
+  case object LibraryDynamic extends NativeBuildTarget("LibraryDynamic")
```
sideeffffect, line 8:
```suggestion
  case object LibraryDynamic extends NativeBuildTarget("libraryDynamic")
```

bleep-model/src/scala/bleep/model/NativeBuildTarget.scala
```diff
@@ -5,24 +5,24 @@ import io.circe.*
...
 
   val All: List[String] =
     List(Application.id, LibraryDynamic.id, LibraryStatic.id)
 
   implicit val encodeNativeBuildTarget: Encoder[NativeBuildTarget] = Encoder.instance {
     case NativeBuildTarget.Application    => Json.obj(("nativeBuildTarget", Json.fromString("application")))
-    case NativeBuildTarget.LibraryDynamic => Json.obj(("nativeBuildTarget", Json.fromString("dynamic")))
-    case NativeBuildTarget.LibraryStatic  => Json.obj(("nativeBuildTarget", Json.fromString("static")))
+    case NativeBuildTarget.LibraryDynamic => Json.obj(("nativeBuildTarget", Json.fromString("libraryDynamic")))
+    case NativeBuildTarget.LibraryStatic  => Json.obj(("nativeBuildTarget", Json.fromString("libraryStatic")))
```
sideeffffect, line 17: One more thing to consider is kebab-case, so `library-static`.
Not sure myself, what do you think?

bleep-model/src/scala/bleep/model/NativeLTO.scala
```diff
@@ -0,0 +1,28 @@
...
+  case object Thin extends NativeLTO("thin")
+  case object Full extends NativeLTO("full")
+
+  val All: List[String] =
+    List(None.id, Thin.id, Full.id)
+
+  implicit val encodeNativeLTO: Encoder[NativeLTO] = Encoder.instance {
+    case NativeLTO.None => Json.obj(("nativeLTO", Json.fromString("none")))
+    case NativeLTO.Thin => Json.obj(("nativeLTO", Json.fromString("thin")))
+    case NativeLTO.Full => Json.obj(("nativeLTO", Json.fromString("full")))
```
sideeffffect, line 17:
```suggestion
    x => Json.obj(("nativeLTO", Json.fromString(x.id)))
```

bleep-model/src/scala/bleep/model/NativeLTO.scala
```diff
@@ -0,0 +1,28 @@
...
+import io.circe.*
+
+sealed abstract class NativeLTO(val id: String)
+object NativeLTO {
+  case object None extends NativeLTO("none")
+  case object Thin extends NativeLTO("thin")
+  case object Full extends NativeLTO("full")
+
+  val All: List[String] =
+    List(None.id, Thin.id, Full.id)
```
sideeffffect, line 12:
```suggestion
  val All: List[NativeLTO] =
    List(None, Thin, Full)
    
  val AllMap: Map[String, NativeLTO] =
    All.map(x => x.id -> x).toMap
```

bleep-model/src/scala/bleep/model/NativeLTO.scala
```diff
@@ -0,0 +1,28 @@
...
+    case NativeLTO.Thin => Json.obj(("nativeLTO", Json.fromString("thin")))
+    case NativeLTO.Full => Json.obj(("nativeLTO", Json.fromString("full")))
+  }
+
+  implicit val decodeNativeLTO: Decoder[NativeLTO] = Decoder.instance { h =>
+    h.get[String]("nativeBuildTarget").flatMap {
+      case "none" => Right(NativeLTO.None)
+      case "thin" => Right(NativeLTO.Thin)
+      case "full" => Right(NativeLTO.Full)
+      case _      => Left(DecodingFailure("NativeLTO", h.history))
```
sideeffffect, line 25:
```suggestion
      s => AllMap.get(s).toRight(DecodingFailure("NativeLTO", h.history))
```

sideeffffect (commented):
I see a lot of repetition of strings like `"thin"`.

Not only is the code unnecessarily verbose, but it's also susceptible to bugs because of typos.

bleep-model/src/scala/bleep/model/NativeBuildTarget.scala
KristianAN: This is the casing scala-native uses, I really have no preference.

sideeffffect: If I remember correctly, it uses `libraryStatic` for the name of the variable (as is the custom in Scala). But it doesn't (yet) have a "string" name for it. But for other things, it uses kebab-case for other strings (e.g. `"release-fast"`).

KristianAN: Then we should just kebab case it.

sideeffffect: Sounds good :+1: 
When I suggested `libraryStatic`, I just copied the name of the variable (because there wasn't anything else to copy).

KristianAN:
> I see a lot of repetition of strings like `"thin"`.
> 
> Not only is the code unnecessarily verbose, but it's also susceptible to bugs because of typos.

100%, a bit of a mess here. Fixed with your suggestions.

oyvindberg (approved):

oyvindberg:
looks good to me! do you think it's ready to merge now?

KristianAN:
Yes,should be good to merge now.

dwalend:
Thanks! I'll try that approach with calico.IOWebApp next.

@oyvindberg - `bleep run` is building something different from `bleep link` , but they should be building the same thing. Do you want a feature request to make this work with the scala-js examples?

KristianAN:
> bleep run is building something different from bleep link , but they should be building the same thing. Do you want a feature request to make this work with the scala-js examples?

I think that this is rather a bug in bloop bsp. I can look into it. bleep link should produce output if a main class is defined. It will not produce output if no main class is definde or there is no JSExport.  I opened an issue here: https://github.com/scalacenter/bloop/issues/2596

PR #446, opened 2024-12-17
Add way to disable warning when using external JVM.

KristianAN:
This does not interfere with the build model, yet provides a simple way of disabling the warning when the JVM version is managed outside of bleep.

Especially useful for Nix users and the likes, where dynamically linked software can not be run without extra effort.


oyvindberg:
sorry, missed this

PR #447, opened 2025-01-02
Support for defining NativeBuildTarget

KristianAN:
When linking SN projects the user can specify to link as an application ,the most common and the default, or the user can specify a dynamic or static library.

This is a new configuration option that has been added to bloop-config for better support for linking without a main class when building SN libraries


oyvindberg:
Nice!

PR #459, opened 2025-03-05
Implement generating dev script on Windows

sake92:
- add windows batch script for dev
- bump coursier since it was causing weird `null` folder to be created.. (fixed in newer versions)


oyvindberg:
thanks! i had a feeling it was going to be easy, i just had no idea about the `%*` syntax :)

Issue #453, opened 2025-02-10
bug: issues with duplicate resources with bloop 2.0.7


KristianAN:
Bloop 2.0.7 introduced copying of resources on compilation (https://github.com/scalacenter/bloop/commit/b7fc29c4d221e3c3ecbcb99ccc0b64d540c2119c). This is causing some weird behaviour in bleep. 

  1. When I am using liquibase and loading the changelog.xml from the resource folder it now complains that it finds two files in the path instead of one.
  2. Changes to files in the resource directory are not applied until a clean -> compile is done. (this might be fixed in 2.0.8 here https://github.com/scalacenter/bloop/commit/306fcaec93f8eca5ce718a8fa034c86105246157)

I'm going to try and create a reproducer and test it with the latest version of bloop to see if it matters for the second point. As for the first point I'm not sure why it picks up both directories (it might just be some insane liquibase thing).

oyvindberg:
gaah, i was so content because it didn't do that!

we're going to change some things then. we're explicitly adding the resource directories to the classpath a place or two - i suppose that's no longer a good idea.

i've actually noticed having resources double up myself, but i hadn't dug into why it happened

KristianAN:
https://github.com/oyvindberg/bleep/blob/master/bleep-model/src/scala/bleep/model/SourceLayout.scala
Is this the right place to start poking around?

oyvindberg:
yes, absolutely. you will find some usage of it in different contexts and we'll have to figure out which ones to keep. 

in particular there is a utility method `fixedClasspath` which computes a new classpath for a project which explicitly adds resource directories - we can at least remove the resource directory in there 

then there is `GenBloopFiles`, where we add resource directories for upstream projects, which means the bloop project file looks like this:
```
    "classpath": [
      "/Users/oyvind/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/ch/epfl/scala/bloop-config_2.13/2.3.1/bloop-config_2.13-2.3.1.jar",
      // ...
      "/Users/oyvind/bleep/.bleep/builds/bsp/.bloop/bleep-model/jvm213/classes",
      "/Users/oyvind/bleep/.bleep/builds/bsp/.bloop/bleep-nosbt/jvm213/classes",
      "/Users/oyvind/bleep/.bleep/generated-resources/bleep-model@jvm213/bleep.scripts.GenerateResources",
      "/Users/oyvind/bleep/bleep-model/src/resources",
[...]

PR #462, opened 2025-05-18
bugfix: Duplicate resources on classpath

KristianAN:
Removed resources from fixedClasspath. No longer observing bugs with duplicate resources on the classpath.

Updated bloop-rifle 2.0.7 -> 2.0.10. Version 2.0.8 contains a fix for resources not being copied on noop, forcing users to clean before building when only changing resources. Bloop now properly copies resources on noop.

I have verified this locally.

The relevant bugfix in bloop: https://github.com/scalacenter/bloop/pull/2579

I have not made any changes in GenBloopFiles just yet as I don't know the implications. However from my quick testing things seem to be working correctly with this small change (mostly because of the bugfix in bloop)


oyvindberg:
nice that you finally could have a go at this! it has been irking me for a while!

i think you'll need to do what i mentioned here as well https://github.com/oyvindberg/bleep/issues/453#issuecomment-2653184268 . otherwise we'll see duplicate resources from upstream projects, i think. i don't think the consequences should be bad, we just replace targetdir + resourcedirs for upstream projects with just targetdir, which has all the resource in it

KristianAN:
It seems like if we just remove the resources from fixedClassPath then plugins such as sbt-native-image is no longer able to find the resources (it failed to find the resources for the native image stuff in META-INF). So i just hacked around it by adding a flag. Maybe there is some more elegant way? In any case I have created PRs in the submodules for the affected liberated plugins.

KristianAN:
The failing tests are expected here right? Because the generated bloop files will look different now that we don't include the resources in the transitive dependencies explicitly.

oyvindberg:
> It seems like if we just remove the resources from fixedClassPath then plugins such as sbt-native-image is no longer able to find the resources (it failed to find the resources for the native image stuff in META-INF). So i just hacked around it by adding a flag. Maybe there is some more elegant way? In any case I have created PRs in the submodules for the affected liberated plugins.

i'm ok with a flag. there may be some dragons here. in particular, there may be old bloop servers with wrong behaviour (according to the new behaviour), or we may run scripts with old versions of bleep code. i'm not entirely sure what you're seeing, but if you need some hacks to get the build through i think that's ok. 

> The failing tests are expected here right? Because the generated bloop files will look different now that we don't include the resources in the transitive dependencies explicitly.

yes. you should be able to just say `bleep test`, it should update some snapshot files which you then check in. it's really just to track all bleep does over time, so changes are wanted

oyvindberg:
so, sorry i've been slow. this looks good now, thank you so much! 

i think the boolean flag flag to include resources should never be set to `true` (it is in the submodules). i also think it won't matter much, so let's touch it up later

PR #463, opened 2025-05-18
ci: Upgrade ubuntu runner to ubuntu 22.04

KristianAN:
Ubuntu 20.04 is now officially deprecated in github actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,7 +53,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


oyvindberg:
can you check that we pick the oldest supported? it's on an old version, because when we compile the native-image it doesnt work for older ubuntus than the one we build on.

KristianAN:
Changed to 22.04 as is the oldest listed here
https://github.com/actions/runner-images

oyvindberg:
diff says 24?

KristianAN:
Might have forgotten to actually push the change

PR #467, opened 2025-07-15
fix: Include scala compiler plugins' transitive dependencies

nafg:
They should be separated with `File.pathSeparator` in a single `-Xplugin` option.


oyvindberg (approved):
looks good. i still have questions how this is supposed to work if there are bincomat problems between transitive dependencies of different compiler plugins - but let's leave that for another day

nafg:
> looks good. i still have questions how this is supposed to work if there are bincomat problems between transitive dependencies of different compiler plugins - but let's leave that for another day

I think the reason plugins are like an array of arrays is it isolates each plugin with its own classloader. That may be an LLM hallucination, but why else would they have this syntax to separate them

PR #490, opened 2026-01-02
Fix flaky --only test discovery (compile and retry)

nafg:
## Problem
Running filtered tests (`--only/--exclude`) intermittently fails with "No tests found in projects" because test discovery can run before compilation populates class data.

## Summary
- retry ScalaTest class discovery after an on-demand compile when filters yield no matching suites
- add a unit test that reproduces the failure and verifies the retry behavior

## Testing
- bleep test bleep-tests --only TestCommandTest


nafg:
Redone (with help from Claude)

PR #491, opened 2026-01-02
Error with mismatched scala versions

nafg:
# Error with mismatched scala versions

## Problem

This PR originally added a failing test demonstrating a runtime error when using ZIO 2.1.24 with Scala 2.13.16 or lower. The error is obscured due to https://github.com/zio/zio/issues/10359, but the real error is:

```
java.lang.NoSuchMethodError: scala.util.hashing.MurmurHash3$.caseClassHash(Product, int, String)
```

This occurs because:
1. The project uses `scala.version = 2.13.16`
2. ZIO 2.1.24 was compiled with Scala 2.13.18
3. ZIO transitively requires scala-library:2.13.18

### Original Questions

> 1. Shouldn't dependency eviction give us the higher Scala version?

**Answer**: Not with Coursier's default behavior. When scala-library is added as an explicit dependency (which bleep was doing via `VersionCombo.libraries()`), Coursier won't upgrade it even with `forceScalaVersion=false`. The "latest wins" strategy only applies to transitive dependencies in conflict. The solution was to stop adding scala-library explicitly for Scala 2.13/3.

> 2. Newer versions of SBT fail in this scenario, requiring you to manually update your `scalaVersion` to not be lower than that of your dependencies. I guess if Bleep would do that too, this wouldn't have stumped me for so long.

**Answer**: Yes! This PR implements full SIP-51 support matching sbt's behavior from [PR #7480](https://github.com/sbt/sbt/pull/7480). Now bleep validates version mismatches and fails with clear error messages.

## Solution: Full SIP-51 Implementation

This PR implements complete SIP-51 (Drop Forwards Binary Compatibility) support for Scala 2.13 and 3. With SIP-51, Scala uses backwards-only binary compatibility: code compiled with 2.13.x works with scala-library 2.13.y where y >= x, but NOT the reverse.

#### 1. Allow scala-library Upgrades (GenBloopFiles.scala)

Filter scala-library from explicit platform dependencies for Scala 2.13/3 so Coursier can freely resolve to the highest version required by transitive dependencies:

```scala
val fromPlatform = versionCombo match {
  case scala: model.VersionCombo.Scala if scala.scalaVersion.is3Or213 =>
    // Only include platform libraries OTHER than scala-library
    versionCombo.libraries(isTest = explodedProject.isTestProject.getOrElse(false))
      .filterNot(dep => dep.organization == Organization("org.scala-lang") &&
                        dep.baseModuleName == ModuleName("scala-library"))
  case _ =>
    // For other versions: include all platform libraries as before
    versionCombo.libraries(isTest = explodedProject.isTestProject.getOrElse(false))
}
```

**Why this works**: By removing scala-library from explicit dependencies, it comes only from transitive dependencies, allowing Coursier to upgrade it when needed.

#### 2. Validate at Compile-Time (commands/Compile.scala)

Detect when resolved scala-library version exceeds scala.version and fail before compilation with actionable error messages:

```scala
def validateSIP51(started: Started, projects: Array[model.CrossProjectName]): Either[BleepException, Unit]
```

This validation:
- Runs at compile-time (not build-load time), so users can run `bleep evicted <project>` to investigate
- Checks only scala-library (the artifact that matters for SIP-51)
- Reports ALL violations across all projects being compiled
- Provides detailed error messages with:
  - Which project has the mismatch
  - Current scala.version vs resolved version
  - Command to investigate: `bleep evicted <project>`
  - Clear fix: upgrade scala.version to the resolved version

#### 3. Fix Scala 3 Macro Classpath (GenBloopFiles.scala)

For Scala 3, macros are loaded from the dependency classpath into the compiler's class loader. Update the compiler's runtime classpath to use the resolved scala-library version:

```scala
if (scalaVersion.is3) {
  val resolvedScalaLibrary = // Extract scala-library from resolved dependencies

  // Replace scala-library.jar in compiler jars with resolved version
  compilerJars.map { jar =>
    if (jar.getFileName.toString.startsWith("scala-library"))
      newLibraryJar
    else
      jar
  }
}
```

This prevents NoSuchMethodException when macros use methods from newer scala-library versions.

### What We Kept

- `forceScalaVersion=false` for Scala 2.13/3 (allows floating library version)
- `scalaVersionOpt` set to scalaVersion (needed for cross-versioned deps like `foo_2.13`)

### What Changed

- scala-library is no longer added as an explicit dependency for Scala 2.13/3 (now only transitive)
- Validation moved from build-load time to compile-time
- Error messages are now project-specific and actionable

## Testing

- **Integration test** (IntegrationTests.scala): Verifies SIP-51 validation with ZIO dependency requiring newer scala-library
- **All integration tests passing**: Including new SIP-51 validation test
- **Snapshot tests updated**: Updated resolve-cache files with new resolution results

## Breaking Changes

None. This is backwards-compatible:
- Only affects Scala 2.13+ and 3
- Scala 2.12 and Java projects unchanged
- Existing projects continue to work; they just get better error messages if there's a mismatch

## References

- **SIP-51 Spec**: https://github.com/scala/scala/pull/11023 (linked from original PR description)
- **SIP-51 Documentation**: https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html
- **sbt Implementation (PR #7480)**: https://github.com/sbt/sbt/pull/7480 - Our implementation matches sbt's approach
- **Backwards-only binary compatibility**: Code compiled with 2.13.x works with library 2.13.y where y >= x

---

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/bleep-core/src/scala/bleep/GenBloopFiles.scala
+++ b/bleep-core/src/scala/bleep/GenBloopFiles.scala
@@ -272,14 +272,28 @@ object GenBloopFiles {
       }
 
     val (resolvedDependencies, resolvedRuntimeDependencies) = {
-      // SIP-51: For Scala 2.13/3, scala-library IS included as an explicit dependency.
-      // This provides a floor constraint - Coursier will resolve to max(scalaVersion, transitive needs).
-      // Combined with forceScalaVersion=false in CoursierResolver, this allows scala-library
-      // to float up when dependencies need a higher version while ensuring it's never lower
-      // than scalaVersion.
+      // SIP-51: scala-library floor constraint handling.
       //
+      // For Scala 2.13: scala-library IS included as an explicit dependency with scalaVersion,
+      // providing a floor constraint. Coursier resolves to max(scalaVersion, transitive needs).
+      //
+      // For Scala 3 before 3.8: scala-library is EXCLUDED from explicit dependencies because
+      // VersionScala hardcodes it to 2.13.18 (via Scala213 constant) which is too high for most
+      // Scala 3 versions. Instead, the correct scala-library version comes from scala3-library's
+      // transitive dependencies.
+      //
+      // For Scala 3.8+: scala-library version matches scalaVersion, so it's included as floor.
+      //
+      // Combined with forceScalaVersion=false and the SameVersion rule in CoursierResolver.
       // See: https://github.com/sbt/sbt/pull/7480
-      val fromPlatform = versionCombo.libraries(isTest = explodedProject.isTestProject.getOrElse(false))
+      val fromPlatform = versionCombo match {
+        case scala: model.VersionCombo.Scala if scala.scalaVersion.is3 && !scala.scalaVersion.is38OrLater =>
+          versionCombo
+            .libraries(isTest = explodedProject.isTestProject.getOrElse(false))
+            .filterNot(dep => dep.organization == Organization("org.scala-lang") && dep.baseModuleName == ModuleName("scala-library"))
+        case _ =>
+          versionCombo.libraries(isTest = explodedProject.isTestProject.getOrElse(false))
+      }
 
       val inherited =
         build.transitiveDependenciesFor(crossName).flatMap { case (_, p) => p.dependencies.values }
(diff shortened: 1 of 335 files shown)


nafg:
P.S. it would be nice if `update-deps` could suggest newer Scala and Bleep versions

oyvindberg:
can you try to point claude at this? make it research how it works in newest sbt, and have bleep do the same? i honestly have no idea about this

PR #494, opened 2026-01-08
Align compile error output more closely with other build tools

mrdziuban:
This changes bleep's error message output to more closely match sbt and mill by making sure it includes:

- The error code, e.g. "E172"
- The error type, e.g. "Type Error"
- The file path, line, and column
- The line of code where the error occurred and the caret pointing to the exact column

It builds on a change I contributed to bloop (https://github.com/scalacenter/bloop/pull/2781) to return the full rendered error message from the compiler when passed the `--show-rendered-message` option.

If these changes look good, then I'd also like to look into making it so that error messages aren't always rendered in all red -- otherwise users need to pass `--no-color` to see syntax highlighting in the compiler error messages. [We talked about this in Discord](https://discord.com/channels/632150470000902164/635669047588945930/1309618081051578398) a little over a year ago, @oyvindberg, and you suggested that we could copy/paste the contents of `LogPatterns` from ryddig into bleep in order to fix it. Is that still the best path forward?

### On current master branch

image

### On current master branch with `--no-color`

image

### On this branch

image

### On this branch with `--no-color`

image

### DIFF
--- a/bleep-core/src/scala/bleep/internal/BspClientDisplayProgress.scala
+++ b/bleep-core/src/scala/bleep/internal/BspClientDisplayProgress.scala
@@ -209,7 +209,7 @@ class BspClientDisplayProgress(
       logger
         .withOptContext("code", Option(d.getCode))
         .withContext("location", location)
-        .log(logLevel, Str(renderBuildTarget(params.getBuildTarget), Str(" "), Str(d.getMessage)))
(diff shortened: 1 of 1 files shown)


sideeffffect:
Shouldn't there be a `\n` before the `-- [E172] Type Error: ...` part?

mrdziuban:
> Shouldn't there be a `\n` before the `-- [E172] Type Error: ...` part?

Yeah, good idea. I added newlines on both sides of the error message to clean things up:

image

PR #567, opened 2026-04-16
Integrate remote cache into BSP compile/test DAG

oyvindberg:
Fixes #546.

## Summary
- Cache pull runs per-project **before** compile inside `makeCompileHandler`. A hit short-circuits the task to `Success`, so downstream DAG nodes start while uncached projects keep compiling.
- Cache push runs in a **detached fiber after** a successful compile — never blocks downstream work or the BSP response.
- New `CompileCacheContext` bundles S3 client, precomputed digests (one `ProjectDigest.computeAll` per build cycle), and an event sink. Becomes `Disabled` when no config, no credentials, or `--no-cache`.
- `RemoteCache` refactored to expose `tryPull`/`tryPush` per-project helpers returning outcome ADTs (never throw). The existing `bleep remote-cache pull/push` commands delegate to them.
- New protocol events: `CachePull{Started,Finished}`, `CachePush{Started,Finished}` with status enums (Hit/Miss/Error/AlreadyCompiled, Success/AlreadyCached/Error).
- Build summary always prints a `Remote cache:` line: either `not configured`, or a breakdown of hits/misses/already-compiled/pushed plus any errors.
- `--no-cache` flag added to `bleep compile`, `bleep link`, `bleep test`, threaded via `params.arguments` into the BSP server.
- Cache failures are warnings, never errors — build always succeeds even when the cache backend is unreachable.

## Test plan
- [x] `bleep compile` (whole build) — green
- [x] `bleep test bleep-tests` — 135/135 pass
- [x] `bleep test bleep-bsp-tests` — 529/529 pass
- [x] `bleep fmt` clean
- [ ] End-to-end against a real S3 bucket: compile, wipe local classes, compile again and confirm cache hits
- [ ] `--no-cache` disables pull/push for the run even with cache configured
- [ ] Network failure during pull surfaces as warning, build proceeds to compile

### DIFF
--- a/bleep-bsp-protocol/src/scala/bleep/bsp/protocol/BleepBspProtocol.scala
+++ b/bleep-bsp-protocol/src/scala/bleep/bsp/protocol/BleepBspProtocol.scala
@@ -495,29 +495,6 @@ object BleepBspProtocol {
       }
     }
 
-    /** Outcome of a remote cache push attempt for a single project */
-    sealed trait CachePushStatus
-    object CachePushStatus {
-      case object Success extends CachePushStatus
-      case object AlreadyCached extends CachePushStatus
-      case class Error(reason: String) extends CachePushStatus
-
-      implicit val errorCodec: Codec[Error] = deriveCodec
-      implicit val encoder: Encoder[CachePushStatus] = Encoder.instance {
-        case Success       => Json.obj("type" -> "Success".asJson)
-        case AlreadyCached => Json.obj("type" -> "AlreadyCached".asJson)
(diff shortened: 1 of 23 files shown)


oyvindberg:
Closing. Decided not to integrate remote cache into the BSP DAG.

Explicit `bleep remote-cache pull` / `bleep remote-cache push` commands are a better design: users control exactly when network I/O happens, there's no ambiguity about when a pull is necessary (no stale freshness heuristics to argue about), and the fail-hard error model stays clean without "when is the cache fast-path correct vs. not" edge cases.

PR #569, opened 2026-04-18
Enable JVM-forking IntegrationTests on CI

oyvindberg:
Stacked on #568.

## Summary
Five \`IntegrationTests\` cases were gated off CI with \`assume(!sys.env.contains(\"CI\"))\`. The root cause wasn't the test logic — it was the forked target JVMs using the default max heap (~25% of system RAM = ~1.75GB on a 7GB CI runner), which put total usage (test JVM + forked target + BSP server in-process + system overhead) over the runner's limit.

Fix: constrain each test's forked JVM explicitly via \`-Xmx256m -Xms32m\`, then drop the skip guard. No logic change, no parallelism change, no test harness change.

## Tests now running on CI

- \`run prefer jvmRuntimeOptions\` — bound added to \`jvmRuntimeOptions\` alongside \`-Dfoo=2\`
- \`run fallback to jvmOptions\` — bound added to \`jvmOptions\` (the fallback path being tested)
- \`scala 3.8.1 with new stdlib\` — bound added to \`jvmRuntimeOptions\`
- \`resource generator\` — target gets 256m, scripts project gets 512m (bleep-core compile needs a bit more)
- \`sourcegen script project that fails to compile\` (new from #568) — no fork in practice (scripts compile fails first), but bound configured defensively

## Budget math

- Test JVM: ~2GB typical
- Forked target: capped at 256MB
- System + overhead: ~500MB
- **Total: ~2.8GB, well under the 7GB CI budget**

## Test plan

- [x] \`bleep test bleep-tests --only IntegrationTests\` — 17/17 pass locally in ~22s
- [ ] CI green on this PR — the point of the PR

### DIFF
--- a/bleep-tests/src/scala/bleep/IntegrationTests.scala
+++ b/bleep-tests/src/scala/bleep/IntegrationTests.scala
@@ -78,14 +78,13 @@ class IntegrationTests extends AnyFunSuite with TripleEqualsSupport {
   }
 
   test("run prefer jvmRuntimeOptions") {
-    assume(!sys.env.contains("CI"), "Skipped on CI: spawns child JVMs that exceed runner memory")
     runTest(
       "run prefer jvmRuntimeOptions",
       """projects:
       a:
         platform:
           name: jvm
-          jvmRuntimeOptions: -Dfoo=2
+          jvmRuntimeOptions: -Xmx256m -Xms32m -Dfoo=2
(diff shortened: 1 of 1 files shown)


oyvindberg:
Folding into #568 — the CI trigger only fires on PRs to master, and these changes are part of the same feature anyway.

PR #658, opened 2026-08-25
Scala.js tests run through TestAdapter and Bridge

### LINKED ISSUE #655: scala.js: bleep test runs no munit suite and no utest suite
## TL;DR. 

`bleep test` runs no Scala.js suite on 1.0.0-M12 because the injected harness
drives the linked output through mangled Scala.js internal names. A munit link declares
none of the names the harness expects, and every munit suite fails at load. A utest suite
loads and then dies inside `TestRunner.runAsync` after a 30 second wait. 

Routing Scala.js tests through the standard `TestAdapter` and `Bridge` repairs both and works for every
sbt-testing framework.

---

`bleep test` on a Scala.js project compiles the project, links it, and discovers the test
suites correctly. Every suite then fails inside the JavaScript harness that
`bleep.bsp.ScalaJsTestRunner` injects. A munit suite fails at load with `Could not load
test module via Reflect`. A utest suite fails after a 30 second wait inside
`utest.TestRunner.runAsync`.

The harness drives the linked output by looking up Scala.js internal symbols under their
mangled JavaScript names. Which mangled names a linked program declares depends on which
code that program reaches, and the names the harness expects are not the names munit
produces.

## Reproducing this takes three projects and three files

`bleep.yaml`

```yaml
$schema: https://raw.githubusercontent.com/oyvindberg/bleep/master/schema.json
$version: 1.0.0-M12
jvm:
  name: corretto:25.0.4.8.1
projects:
  jstest:
    dependencies:
    - org.scalameta::munit:1.3.5
    isTestProject: true
    platform:
      name: js
      jsVersion: 1.22.0
      jsNodeVersion: 24.18.0
      jsKind: none
    scala:
      version: 3.8.4
  jvmtest:
    dependencies:
    - org.scalameta::munit:1.3.5
    isTestProject: true
    platform:
      name: jvm
    scala:
      version: 3.8.4
  utesttest:
    dependencies:
    - com.lihaoyi::utest:0.9.1
    isTestProject: true
    platform:
      name: js
      jsVersion: 1.22.0
      jsNodeVersion: 24.18.0
      jsKind: none
    scala:
      version: 3.8.4
```

`jstest/src/scala/repro/ClassSpec.scala`, copied byte for byte to
`jvmtest/src/scala/repro/ClassSpec.scala`

```scala
package repro

class ClassSpec extends munit.FunSuite:

  test("one plus one is two") {
    assertEquals(1 + 1, 2)
  }
```

`utesttest/src/scala/repro/UtestSpec.scala`

```scala
package repro

import utest._

object UtestSpec extends TestSuite:
  val tests = Tests {
    test("two plus two is four") {
      assert(2 + 2 == 4)
    }
  }
```

`bleep test jvmtest` passes.

```
[info ]   Tests: 1 passed, 0 failed
[info ]   Suites: 1 total
```

`bleep test jstest` fails on the identical source.

```
[info ] 🔍 discovered 1 test suites (total: 1) [project => jstest]
[info ] FAILED repro.ClassSpec: 0 passed, 1 failed, 0 skipped (511 ms)

[info ] ! jstest / repro.ClassSpec
[info ]   | Suite reported 1 failure(s) but no individual test results were captured
[info ]   Output:
[info ]   | Could not load test module via Reflect: repro.ClassSpec$
```

`bleep test utesttest` fails after 30 seconds.

```
[info ] 🔍 discovered 1 test suites (total: 1) [project => utesttest]
[info ] NO TESTS repro.UtestSpec: discovered but executed 0 tests (30412 ms)

[info ] ! utesttest / repro.UtestSpec
[info ]   | Suite repro.UtestSpec was discovered but executed 0 tests
[info ]   Output:
[info ]   | scala.scalajs.js.JavaScriptException: TypeError: $n(...).apply__O__O is not a function
[info ]   |   at scala.concurrent.impl.Promise$Transformation.run(.../main.js:54210:17)
```

I expected the Scala.js runs to pass, matching the JVM run and matching what sbt and mill
produce through the standard Scala.js test adapter.

## A munit suite fails because the harness looks up a Reflect module the linker did not emit

`bleep-bsp/src/scala/bleep/bsp/ScalaJsTestRunner.scala:411` looks the reflection entry
point up under one fixed mangled name.

```js
const ReflectGetter = sandbox[dollar + 'm_Lorg_portablescala_reflect_Reflect' + dollar];
```

The Scala.js linker declares a module accessor only for a module the linked program
reaches. utest's framework code calls `org.portablescala.reflect.Reflect`. munit calls
`scala.scalajs.reflect.Reflect` from `munit.internal.PlatformCompat`. Counting the two
names with `grep -c` on the two linked `main.js` files this reproduction produced gives:

| linked output | `$m_Lorg_portablescala_reflect_Reflect$` | `$m_sjs_reflect_Reflect$` |
| --- | --- | --- |
| `utesttest` | 3 | 9 |
| `jstest` | 0 | 10 |

`ReflectGetter` is `undefined` for the munit project. `Reflect` becomes `null`. The
loader at line 436 then reports `Could not load test module via Reflect` for every suite
in the project, whatever the suite is. Replacing the class in the reproduction with a
Scala `object` changes nothing, even though the linked output registers that object
through `registerLoadableModuleClass("repro.ObjectSpec$", ...)` and invokes the
registration before the bridge starts.

## The loader needs a class branch for munit

`ScalaJsTestRunner.scala:422` appends a dollar sign to the suite name. Line 426 calls
`lookupLoadableModuleClass` and nothing else. munit's `Framework` declares
`SubclassFingerprint(superclassName = "munit.Suite", isModule = false,
requireNoArgConstructor = true)`, which makes a munit suite a class rather than an
object. The JVM run above confirms that reading. munit's own `PlatformCompat.newRunner`
calls `Reflect.lookupInstantiatableClass(taskDef.fullyQualifiedName())` with no dollar
sign, which line 8001 of the linked `main.js` for `jstest` shows. A working Reflect
lookup alone would still leave `class X extends munit.FunSuite` unable to load, because
the harness declares no `lookupInstantiatableClass` plus `newInstance` path.

## The runner drives utest only

`ScalaJsTestRunner.scala:443` fetches `tests__Lutest_Tests` from the loaded suite. Line
445 reports `No tests__Lutest_Tests method found on suite` when that method is absent.
`runTests` declares no munit execution path. The discovery script's `isTestSuite` already
detects `prototype.munitTests`, and the runner has no branch matching that detection. A
munit suite that loaded successfully would fail here instead.

## The utest suite fails as well

The `TypeError` above comes from inside `utest.TestRunner.runAsync`, which the harness
calls at `ScalaJsTestRunner.scala:458` onward with hand-built `AnonFunction1` and
`AnonFunction2` instances. I did not root-cause that failure. The 30 second wait before
the run gives up comes from the hard-coded `setTimeout(() => resolve(), 30000)` at line
533.

## The linked output already declares the standard bridge

Every linked test `main.js` this reproduction produced ends on the standard entry point.

```js
$sct_Lmunit_Framework__stinit__();
$sct_Lrepro_ClassSpec__stinit__();
$s_Lorg_scalajs_testing_bridge_Bridge__start__V();
```

The harness deletes that call with a regular expression and takes the run over.

## Proposed change

Drive Scala.js tests through `org.scalajs.testing.adapter.TestAdapter` on the JVM side
and `org.scalajs.testing.bridge.Bridge` on the JavaScript side, which is the path sbt and
mill take. `TestAdapter` speaks the `sbt.testing` interfaces that bleep's JVM-side
discovery already uses. It works for munit, scalatest, specs2, and utest without a code
path per framework, and it removes every dependence on a mangled name that the linker
declares only when the program happens to reach that code. `TestAdapter` also accepts a
`JSEnv`, which gives a later hook for `JSDOMNodeJSEnv` when a suite touches the DOM.

Two cheaper repairs would fix munit alone. The loader could fall back to
`lookupInstantiatableClass(suiteName)` plus `newInstance`, and `runTests` could grow a
munit branch. Both keep the mangled-name coupling that produced this defect.

You said in discussion #654 that a pull request for the `TestAdapter` route is welcome. I
am ready to write it against this issue.

## Relation to #233

[#233](https://github.com/oyvindberg/bleep/issues/233) reports Scala.js tests hanging on
0.0.1-M22 and Scala Native frameworks going undetected. The harness described here
replaced the code that #233 reported on, and the version, the symptoms, and the root
cause are all different. Close #233 as covered by this issue if you prefer to track one.

## Environment

- bleep 1.0.0-M12, with the harness source unchanged on master at `b0cea39f`
- macOS 26.5.2 on arm64
- JVM `corretto:25.0.4.8.1`, managed by bleep
- Scala 3.8.4, Scala.js 1.22.0, node 24.18.0 managed by bleep
- munit 1.3.5, utest 0.9.1

dwalend:
**TL;DR.** `bleep test` ran no Scala.js suite because the JavaScript harness
that `ScalaJsTestRunner` injected reached the linked module through mangled
Scala.js internal names. This huge PR replaces that harness and runs the linked
module through `org.scalajs.testing.adapter.TestAdapter`, which speaks the
sbt-testing protocol to the `Bridge` that bleep already writes into every test
module. Every sbt-testing framework implements the `sbt.testing.Framework` that
adapter returns, which puts munit, utest, and ScalaTest on one flow-of-control.
`SbtTestDriver` runs that framework for Scala.js. `SbtTestDriver` also runs the
framework for Scala Native, which already reached its framework through its own
`TestAdapter`.

Closes #655.

## The injected harness could not load a munit suite

`bleep test` on a Scala.js project compiled the project, linked it, and
discovered its suites. Every suite then failed inside the JavaScript harness
that `ScalaJsTestRunner` wrote to a temporary file. A munit suite failed at load
with `Could not load test module via Reflect`. A utest suite loaded and then
died inside `utest.TestRunner.runAsync` after a thirty second wait.

That harness reached the linked program through Scala.js internal symbols under
their mangled JavaScript names. A linked program declares whichever mangled
names the code it reaches produces. munit produces different names from the
names the harness expected.

The harness also invented a text protocol for its own output. It printed each
event as a prefixed line of fields joined by a vertical bar. `ScalaJsTestRunner`
parsed those lines back into events.

## Scala.js publishes an adapter for this work

`org.scalajs.testing.adapter.TestAdapter` starts node. The adapter speaks the
sbt-testing protocol to `org.scalajs.testing.bridge.Bridge`, which the Scala.js
linker already puts into every linked test module. The adapter returns an
`sbt.testing.Framework`. Every sbt-testing framework implements that interface.
munit, utest, and ScalaTest therefore each run without bleep knowing anything
particular about them.

`CompilerResolver.getScalaJsTestAdapter` resolves
`scalajs-sbt-test-adapter_2.13` and `scalajs-env-nodejs_2.13` into an isolated
classloader (`bleep-bsp/src/scala/bleep/analysis/CompilerResolver.scala:597`).
One Coursier fetch takes both artifacts because Coursier reconciles a shared
dependency that the two artifacts request at two versions. Two fetches would put
both versions on the classpath. `resolveDepsInOneFetch` does that work at
`bleep-bsp/src/scala/bleep/analysis/CompilerResolver.scala:400`.

`ScalaJsTestRunner.runTestsViaAdapter` replaces the harness
(`bleep-bsp/src/scala/bleep/bsp/ScalaJsTestRunner.scala:42`). That file went
from 573 lines to 259.

## SbtTestDriver runs a framework for both adapter platforms

Scala Native reaches its `sbt.testing.Framework` through
`scala.scalanative.testinterface.adapter.TestAdapter`. Scala.js reaches its
framework through the Scala.js adapter. Running the framework afterward is
identical work. `SbtTestDriver.runFramework` does that work once for both
platforms (`bleep-bsp/src/scala/bleep/bsp/SbtTestDriver.scala:34`).
`ScalaNativeTestRunner` lost its copy.

Two details in the driver earn a comment. `runner.done()` is the barrier that
flushes events. A Scala.js adapter reports its events over a socket, and
`Task.execute` can return before the last event arrives.

A framework puts whatever it likes into `Event.fullyQualifiedName`. munit puts
the suite name and the test name joined by a dot, which is not a suite name. The
driver takes the suite name from the `TaskDef` the framework handed back
(`bleep-bsp/src/scala/bleep/bsp/SbtTestDriver.scala:53`).

## AlienValue keeps an alien object beside its classloader

The adapter classloader brings its own `scala-library`. The adapter's methods
take and return that library's `List`, `Map`, and `Option`. Those three classes
differ from bleep's own three classes of the matching names. Every call into the
adapter therefore goes through reflection.

`AlienList`, `AlienMap`, and `AlienOption` each pair the alien object with the
classloader that built it (`bleep-bsp/src/scala/bleep/bsp/AlienValue.scala:10`).
The raw reference stays `private[bsp]`, which keeps loose reflection beside the
test adapters. Both `ScalaJsTestRunner` and `ScalaNativeTestRunner` replaced
their conversion helpers with these three types.

## The DAG links a test project once

`runScalaJsTestSuite` used to run its own `LinkTask` for each suite. The DAG
already links the project before any of its suites runs. The method now finds
the module that link wrote through `LinkExecutor.locateLinkedJs`
(`bleep-bsp/src/scala/bleep/bsp/LinkExecutor.scala:111`), and throws
`NoLinkedScalaJsModuleException` when no link has written a module
(`bleep-bsp/src/scala/bleep/bsp/MultiWorkspaceBspServer.scala:4367`). A project
with five suites now links once rather than five times.

## Cancellation stops node rather than closing a socket

`TestAdapter.close` closes the adapter's socket to node. A suite that spins
never returns to node's event loop. Such a suite never notices the socket
closing. The thread waiting for that suite's reply does not return either, and
the node process outlives the run.

`RecordingJsEnv` wraps the `JSEnv` the adapter uses in a proxy that keeps every
`JSRun` the adapter starts
(`bleep-bsp/src/scala/bleep/bsp/ScalaJsTestRunner.scala:127`). Cancellation
closes each of those runs, which reaches the node process through
`ExternalJSRun.close` and `destroyForcibly`. `startWithCom` returns a `ComRun`
that wraps the process-level run. Closing the process-level run first releases
the monitor `ComRun.close` needs.

## InMemoryPipe replaces PipedInputStream in the in-process server

`java.io.PipedInputStream` records the thread that wrote most recently. Its
`read` throws `Write end dead` once that thread has exited and the buffer has
run empty, even while other threads keep writing. Cats Effect retires the
carrier thread of a blocking region. A BSP notification sent from such a region
leaves a retired thread recorded, and the next read fails.

`InMemoryPipe` records no thread
(`bleep-bsp/src/scala/bleep/bsp/InMemoryPipe.scala:15`). A read blocks until a
write arrives or until the sink closes. `InProcessBspServer` now builds one pipe
per direction.

## New suites cover the code this branch adds

`ScalaJsTestAdapterIntegrationTest` compiles and links real munit and utest
sources, then runs them through the adapter. It covers a passing test and a
failing test for each framework, an ESModule link, environment variables
reaching the node process, a kill signal that fired before the run started,
cancellation of a suite that never returns, and three runs in parallel. Two of
its assertions matter most. Cancellation returns within a bounded time.
Cancellation leaves no node process behind.

`SbtTestDriverTest` runs a fake `sbt.testing.Framework` and checks the counts,
the durations, the failure messages, and the walk over nested tasks.
`AlienValueTest` builds a second classloader and checks each conversion.
`InMemoryPipeTest` covers the wrap-around, the blocking write, and both close
directions.

`EdgeCaseIntegrationTest` and `TimeoutAndResourceTest` lost thirteen Scala.js
tests. Those tests fed malformed and truncated harness output to the parser this
branch deletes. Their parallel-run test moved into
`ScalaJsTestAdapterIntegrationTest`.

`ScalaJsTestRunner.NodeEnvironment.JSDOM` is gone. `runScalaJsTestSuite` passed
`NodeEnvironment.Node` at every call, so no build ever selected the JSDOM case.

## I ran both suites and a real Scala.js build

`bleep test bleep-bsp-tests` reports 727 passed and 1 failed. The failure is the
Boehm GC test in `ScalaNativeAdvancedLinkIntegrationTest`, which fails in the
LLVM linker on my machine and touches no code this branch changes. `bleep test
bleep-tests` reports 398 passed.

A private repository of mine builds a Scala.js browser application and a
Scala.js test project. Against a locally published build of this branch,
`clean-all`, `link-all`, and `test` each succeeded. That test run reported 592
passed and 0 failed across 85 suites. Its Scala.js suite ran three tests through
the adapter and left no node process behind.

## A reviewer should focus on four things

- The reflection in `AlienValue` and in `ScalaJsTestRunner.JsTestAdapter`, which
  no compiler checks.
- The `ComRun` field name `org$scalajs$jsenv$nodejs$ComRun$$run`, which a
  Scala.js release could rename.
- The `catch { case _: Exception => () }` around `runner.done()` at
  `bleep-bsp/src/scala/bleep/bsp/SbtTestDriver.scala:130`. A Scala Native
  adapter throws there when the native process's RPC handler declares no
  `done` opcode. A Scala.js adapter runs the call to completion. The empty
  catch is the one place this branch swallows a failure.
- `CompilerResolver.ScalaJsNodeEnvVersion`, which pins `scalajs-env-nodejs` at
  1.6.0 rather than deriving it from the project's Scala.js version.


dwalend:
I see you did a lot of this in parallel since I posted the bug and promised a fix. 

I tested current `master` (`9f4da75a`, which includes `072577f4`) against a real
Scala.js project. The test run fails the one-real-test I put in a Scala.js suite.

SbtTestingBridge.ScalaColl.toMap throws on the second environment variable

```
❌ Test net.walend.duck.front.maplibregl.GeoJsonJsSpec failed:
java.lang.IllegalArgumentException: object of type scala.collection.immutable.Map$Map1
  is not an instance of scala.collection.immutable.Map$EmptyMap$
```

Do you have a pending fix for that?

oyvindberg:
Yeah sorry! I misread, and just pointed my agent at the report. 

I missed running the last full harness it built before merge, so that big appeared instantly. I'll see if there is more to bring in from your pr, and fix it up tomorrow :)

Thanks anyway!

dwalend:
I've got maybe another 20 things to fix - some big, some small. Want me to post issues for the whole lot of them?

oyvindberg:
Please do :)

oyvindberg:
rescued some parts into #659

